PR #13, opened 2019-10-03
Added alert dialogs for Capture, Mask and Metadata activity.

davidegavio:


app/src/main/java/net/aiscope/gdd_app/ui/capture/CaptureImageActivity.kt
```diff
@@ -34,6 +38,30 @@ class CaptureImageActivity : AppCompatActivity(), CaptureImageView {
...
+                        dialog.dismiss()
+                    }
+                }
+                // Set other dialog properties
+                builder.setMessage("Are you sure you want to exit? Your current data will be lost")
+                builder.setTitle("Warning!")
+                // Create the AlertDialog
+                builder.create()
+            }
+            alertDialog!!.show()
```
eduardb: Is there a way not to use the double-bang operator here? Since it is considered an antipattern usually.

app/src/main/java/net/aiscope/gdd_app/ui/capture/CaptureImageActivity.kt
```diff
@@ -34,6 +38,30 @@ class CaptureImageActivity : AppCompatActivity(), CaptureImageView {
...
+                        startActivity(intent)
+                    }
+                    setNegativeButton(R.string.cancel
+                    ) { dialog, _ ->
+                        // User cancelled the dialog
+                        dialog.dismiss()
+                    }
+                }
+                // Set other dialog properties
+                builder.setMessage("Are you sure you want to exit? Your current data will be lost")
```
eduardb: Could you please extract all strings to the resources?

app/src/main/java/net/aiscope/gdd_app/ui/capture/CaptureImageActivity.kt
```diff
@@ -34,6 +38,30 @@ class CaptureImageActivity : AppCompatActivity(), CaptureImageView {
...
         setContentView(R.layout.activity_capture_image)
         setSupportActionBar(toolbar)
+        toolbar.setNavigationOnClickListener {
+            val alertDialog: AlertDialog? = this.let {
+                val builder = AlertDialog.Builder(it, R.style.MyAlertDialogStyle)
+                builder.apply {
+                    setPositiveButton(R.string.ok
+                    ) { _, _ ->
+                        // User clicked OK button
+                        val intent = Intent(context,  MainActivity::class.java)
```
eduardb: I think that it's also needed to clear the activity stack here, otherwise, the user will be able to return back to this activity by using the back button, which should not be allowed. See the `goToHome()` method in the `MetadataActivity`.

eduardb:
Thanks for your contribution, @davidegavio!

Issue #16, opened 2019-10-09
Add lint and fix issues


eduardb:
What the title says.

VivekSingh777:
Can I grab this issue?

eduardb:
Hi @VivekSingh777! Yes, please, go for it :)

eduardb:
Hi @VivekSingh777! Do you have any updates on this? Take also into consideration breaking it down in several iterations, doing everything in one go might be too much of an effort.

VivekSingh777:
Sorry I don't required resources to install this on my local. Can u reassign to some one else.

eduardb:
Sure, no problem. Thanks for showing your interest in the first place :)

subbramanil:
@eduardb Let me take a stab at it. 

Bit confused about 'Add lint'.

From Android Studio, I see the below results. 

[Lint Results](https://www.dropbox.com/s/yrr507odsh71vm9/Screen%20Shot%202019-10-16%20at%2012.20.11%20AM.png?dl=0)

I can't promise that I can fix everything, but I'm confident about most of it. 

Let me know

eduardb:
Hi @subbramanil! Sorry for getting back this late to you.
Yes, you are right, Lint is already running, that was a mistake from my part. Maybe what else could be done would be making the build fail if new lint errors are added.
Feel free to fix as much as you can, and then just create a baseline for the rest. We like to iterate :) 
Cheers!

PR #20, opened 2019-10-23
hotfix: Addresses Lint Warnings

subbramanil:
- Cleans up unused import statements
- Applies appropriate scope qualifiers (i.e., private) for variables & functions
- Moves hard-coded strings to String.xml & consts as required
- Cleans up Regular expressions by removing unnecessary character escapes

**Note:**
1. Almost all of the fixes were applied automatically through Android Studio's Auto-Suggestions which made sense & appropriate. None of the changes should affect the functionality of the app. No files were removed, although there were unused files & references.
2. Reports character escapes that are replaceable with the unescaped character without a change in meaning. Note that inside the square brackets of a character class, many escapes are unnecessary that would be necessary outside of a character class. For example the regex [\.] is identical to [.]
3. None of the changes should affect any functionality. But in case of any doubt, any change can be removed, analyzed & applied on another iteration.

Reference: #16


app/src/main/java/net/aiscope/gdd_app/ui/metadata/ListItem.kt
```diff
@@ -3,6 +3,6 @@ package net.aiscope.gdd_app.ui.metadata
 data class ListItem(val id: Long, val name: String) {
 
     override fun toString(): String {
-        return name.orEmpty()
+        return name
```
eduardb: Haha, nice one

app/src/main/java/net/aiscope/gdd_app/ui/mask/customview/MaskCustomView.kt
```diff
@@ -146,9 +147,7 @@ class MaskCustomView @JvmOverloads constructor(
             mY = y
 
             // TODO: improve performance, see if we can to canvas on touch-up
-            maskCanvas?.let {
-                it.drawPath(path, maskPaint)
-            }
+            maskCanvas?.drawPath(path, maskPaint)
```
eduardb: 💯

app/src/main/res/layout/activity_new_health_facility.xml
```diff
@@ -9,33 +9,35 @@
     <TextView
         android:id="@+id/text_health_facility_name_hint"
         style="@style/TextHint"
-        android:text="@string/text_health_facility_name_hint"
+        android:labelFor="@id/text_health_facility_name_field"
```
eduardb: I think that here the `android:text` should be kept as well, along with the new attribute.

subbramanil: Yes, You are correct.

app/src/main/res/layout/activity_new_health_facility.xml
```diff
@@ -9,33 +9,35 @@
...
 
     <EditText
         android:id="@+id/text_health_facility_name_field"
         android:layout_width="0dp"
         android:layout_height="wrap_content"
         android:layout_marginTop="18dp"
         android:ems="10"
-        android:hint="Hospital name"
         android:inputType="textPersonName"
+        android:text="@string/text_health_facility_name_hint"
```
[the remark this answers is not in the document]
subbramanil: Yes, somehow I confused between the two attributes. :D

subbramanil:
Updated the PR with the fixes.

PR #23, opened 2020-04-02
Default the Sample metadata to the settings of the user's last created Sample

CeciliaHwang:
Story: Metadata - keep last selected by user
AC: When the microscopist accesses the app the second time, the Smear Type, Species and Stage that the user selected the last time are shown as default values.

Added a createdOn field to the Sample in order to determine which was the last Sample created. This only looks at Samples that are completed (not incomplete). Since createdOn is new this functionality will only work once a Sample has been created (and completed) with the updated code.

### DIFF
--- a/app/src/main/java/net/aiscope/gdd_app/dagger/AppComponent.kt
+++ b/app/src/main/java/net/aiscope/gdd_app/dagger/AppComponent.kt
@@ -4,7 +4,6 @@ import dagger.BindsInstance
 import dagger.Component
 import dagger.android.AndroidInjectionModule
 import net.aiscope.gdd_app.application.GddApplication
-import net.aiscope.gdd_app.ui.metadata.MetadataMapper
 import javax.inject.Singleton
 
 @Singleton
--- a/app/src/main/java/net/aiscope/gdd_app/dagger/AppComponent.kt
+++ b/app/src/main/java/net/aiscope/gdd_app/dagger/AppComponent.kt
@@ -4,6 +4,7 @@ import dagger.BindsInstance
 import dagger.Component
 import dagger.android.AndroidInjectionModule
 import net.aiscope.gdd_app.application.GddApplication
+import net.aiscope.gdd_app.ui.metadata.MetadataMapper
 import javax.inject.Singleton
 
 @Singleton
--- a/app/src/main/java/net/aiscope/gdd_app/repository/SampleRepositorySharedPreference.kt
+++ b/app/src/main/java/net/aiscope/gdd_app/repository/SampleRepositorySharedPreference.kt
@@ -58,6 +58,6 @@ class SampleRepositorySharedPreference @Inject constructor(
                         && s.microscopist == facility.microscopist
                         && s.createdOn != null
                         && s.status != Status.Incomplete}
-            .sortedWith(compareBy {it.createdOn}).lastOrNull()
+            .sortedBy { it.createdOn }.lastOrNull()
     }
 }
--- a/app/src/main/java/net/aiscope/gdd_app/repository/SampleRepositorySharedPreference.kt
+++ b/app/src/main/java/net/aiscope/gdd_app/repository/SampleRepositorySharedPreference.kt
@@ -3,6 +3,8 @@ package net.aiscope.gdd_app.repository
 import com.github.salomonbrys.kotson.fromJson
 import com.google.gson.Gson
 import net.aiscope.gdd_app.model.Sample
+import net.aiscope.gdd_app.model.Status
+import java.util.Calendar
 import javax.inject.Inject
 
 class SampleRepositorySharedPreference @Inject constructor(
@@ -21,7 +23,7 @@ class SampleRepositorySharedPreference @Inject constructor(
     override suspend fun create(): Sample {
         val uuid = uuid.generateUUID()
         val facility = healthFacilityRepository.load()
-        val sample = Sample(uuid, facility.id, facility.microscopist)
+        val sample = Sample(uuid, facility.id, facility.microscopist, createdOn = Calendar.getInstance())
 
         currentSample = sample
         return sample
@@ -46,5 +48,16 @@ class SampleRepositorySharedPreference @Inject constructor(
         return jsons.map {
             gson.fromJson<SampleDto>(it).toDomain()
         }.toList()
+    }
+
+    override suspend fun last(): Sample? {
+        val allStores = all()
+        val facility = healthFacilityRepository.load()
+        return allStores
+            .filter{s -> s.healthFacility == facility.id
+                        && s.microscopist == facility.microscopist
+                        && s.createdOn != null
+                        && s.status != Status.Incomplete}
+            .sortedWith(compareBy {it.createdOn}).lastOrNull()
     }
 }
--- a/app/src/main/java/net/aiscope/gdd_app/repository/SampleRepositorySharedPreference.kt
+++ b/app/src/main/java/net/aiscope/gdd_app/repository/SampleRepositorySharedPreference.kt
@@ -52,11 +52,8 @@ class SampleRepositorySharedPreference @Inject constructor(
 
     override suspend fun last(): Sample? {
         val allStores = all()
-        val facility = healthFacilityRepository.load()
         return allStores
-            .filter{s -> s.healthFacility == facility.id
-                        && s.microscopist == facility.microscopist
-                        && s.createdOn != null
+            .filter{s -> s.createdOn != null
                         && s.status != Status.Incomplete}
             .sortedBy { it.createdOn }.lastOrNull()
     }
--- a/app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataActivity.kt
+++ b/app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataActivity.kt
@@ -61,7 +61,7 @@ class MetadataActivity : AppCompatActivity() , MetadataView, CaptureFlow {
         var index = -1
         for (i in 0 until spinner.adapter.count){
             if (spinner.adapter.getItem(i) == value) {
-                index = 1
+                index = i
                 break;
             }
         }
--- a/app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataActivity.kt
+++ b/app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataActivity.kt
@@ -58,15 +58,9 @@ class MetadataActivity : AppCompatActivity() , MetadataView, CaptureFlow {
     }
 
     private fun selectSpinnerValue(spinner: AbsSpinner, value: String) {
-        var index = -1
-        for (i in 0 until spinner.adapter.count){
-            if (spinner.adapter.getItem(i) == value) {
-                index = i
-                break;
-            }
-        }
-        if (index > -1)
-            spinner.setSelection(index)
+        (0 until spinner.adapter.count)
+            .firstOrNull { spinner.adapter.getItem(it) == value }
+            ?.let { spinner.setSelection(it) }
     }
 
     override fun onDestroy() {
--- a/app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataActivity.kt
+++ b/app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataActivity.kt
@@ -2,6 +2,7 @@ package net.aiscope.gdd_app.ui.metadata
 
 import android.content.Intent
 import android.os.Bundle
+import android.widget.AbsSpinner
 import android.widget.Toast
 import androidx.appcompat.app.AppCompatActivity
 import androidx.recyclerview.widget.LinearLayoutManager
@@ -13,9 +14,6 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
 import net.aiscope.gdd_app.R
-import net.aiscope.gdd_app.model.MalariaSpecies
-import net.aiscope.gdd_app.model.MalariaStage
-import net.aiscope.gdd_app.model.SmearType
 import net.aiscope.gdd_app.ui.CaptureFlow
 import net.aiscope.gdd_app.ui.attachCaptureFlowToolbar
 import net.aiscope.gdd_app.ui.capture.CaptureImageActivity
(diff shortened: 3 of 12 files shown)


app/src/main/java/net/aiscope/gdd_app/repository/SampleRepositorySharedPreference.kt
```diff
@@ -47,4 +49,15 @@ class SampleRepositorySharedPreference @Inject constructor(
...
+
+    override suspend fun last(): Sample? {
+        val allStores = all()
+        val facility = healthFacilityRepository.load()
+        return allStores
+            .filter{s -> s.healthFacility == facility.id
+                        && s.microscopist == facility.microscopist
+                        && s.createdOn != null
+                        && s.status != Status.Incomplete}
+            .sortedWith(compareBy {it.createdOn}).lastOrNull()
```
eduardb, line 61: Could be written shorter
```suggestion
            .sortedBy { it.createdOn }.lastOrNull()
```

app/src/main/java/net/aiscope/gdd_app/repository/SampleRepositorySharedPreference.kt
```diff
@@ -21,7 +23,7 @@ class SampleRepositorySharedPreference @Inject constructor(
     override suspend fun create(): Sample {
         val uuid = uuid.generateUUID()
         val facility = healthFacilityRepository.load()
-        val sample = Sample(uuid, facility.id, facility.microscopist)
+        val sample = Sample(uuid, facility.id, facility.microscopist, createdOn = Calendar.getInstance())
```
eduardb, line 26: 👍 
In the future, when Android will support them, we could use `java.time` classes. But for now `Calendar` seems to be a good choice.

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataActivity.kt
```diff
@@ -51,44 +49,24 @@ class MetadataActivity : AppCompatActivity() , MetadataView, CaptureFlow {
...
-                presenter.save(selectedSmearType(), selectedSpecies(), selectedStage())
+                presenter.save(
+                    MetadataMapper.getSmearType(metadata_section_smear_type_radio_group.checkedRadioButtonId),
+                    MetadataMapper.getSpecies(baseContext, metadata_species_spinner.selectedItem),
+                    MetadataMapper.getStage(baseContext, metadata_stage_spinner.selectedItem))
             }
         }
     }
 
-    private fun selectedSmearType(): SmearType {
```
eduardb, line 59: Glad to see this logic extracted 👏

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataMapper.kt
```diff
@@ -0,0 +1,80 @@
...
+
+import android.content.Context
+import android.content.res.Resources
+import net.aiscope.gdd_app.R
+import net.aiscope.gdd_app.model.MalariaSpecies
+import net.aiscope.gdd_app.model.MalariaStage
+import net.aiscope.gdd_app.model.SmearType
+
+class MetadataMapper {
+    private constructor () { /* make constructor private for utility class */ }
```
eduardb, line 11: While I think that this is a very good pattern in Java, in Kotlin doesn't seem to be idiomatic. Since you already have a companion object, a singleton instance will be created, so why don't make the `MetadataMapper` an `object` (thus a singleton)? In this way, the instance of this singleton could be injected into `MetadataActivity`, avoiding the static methods calls (which in general cannot be stubbed in tests) – given that the `@JvmStatic` annotations are removed from the methods. We don't have Java in the codebase, so I would say that this is premature optimization.

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataMapper.kt
```diff
@@ -0,0 +1,80 @@
+package net.aiscope.gdd_app.ui.metadata
+
+import android.content.Context
+import android.content.res.Resources
+import net.aiscope.gdd_app.R
+import net.aiscope.gdd_app.model.MalariaSpecies
+import net.aiscope.gdd_app.model.MalariaStage
+import net.aiscope.gdd_app.model.SmearType
+
+class MetadataMapper {
```
eduardb, line 10: Like I was saying above, it's great that this logic is extracted here. Do you think that it could also be unit-tested?

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataPresenter.kt
```diff
@@ -16,7 +16,8 @@ data class ViewStateModel(
     val disease: String,
     val images: List<File>,
     val options: List<FieldOption>,
-    val required: Boolean = true
+    val required: Boolean = true,
+    val sampleMetadata: SampleMetadata? = null
```
eduardb, line 20: Do you think that instead of adding the domain model as a field of the view state model, it would be better to add the individual fields that we need? Also, maybe we can do the mapping with the `MetadataMapper` here, instead of the activity, just a thought.

app/src/test/java/net/aiscope/gdd_app/presentation/MetadataPresenterTest.kt
```diff
@@ -86,4 +88,31 @@ class MetadataPresenterTest {
             assertEquals(it.metadata, expected)
         })
     }
+
+    @Test
+    fun shouldDefaultToLastMetadata() = coroutinesTestRule.runBlockingTest{
```
eduardb, line 93: 💯

app/src/test/java/net/aiscope/gdd_app/repository/SampleRepositorySharedPreferenceTest.kt
```diff
@@ -112,4 +114,27 @@ class SampleRepositorySharedPreferenceTest {
 
         assert(samples.size == 2)
     }
+
+    @Test
+    fun `should give last sample`() = coroutinesTestRule.runBlockingTest {
```
eduardb, line 119: 👍

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataActivity.kt
```diff
@@ -51,44 +49,24 @@ class MetadataActivity : AppCompatActivity() , MetadataView, CaptureFlow {
...
-            getString(R.string.malaria_stage_schizont) -> MalariaStage.SCHIZONT
-            getString(R.string.malaria_stage_gametocyte) -> MalariaStage.GAMETOCYTE
-            else -> throw IllegalStateException(
-                "${metadata_species_spinner.selectedItem} stage is unknown"
-            )
+    private fun selectSpinnerValue(spinner: AbsSpinner, value: String) {
+        var index = -1
+        for (i in 0 until spinner.adapter.count){
+            if (spinner.adapter.getItem(i) == value) {
+                index = 1
```
eduardb, line 64: Ha, I was under the impression that something was strange, and I found it. I have made the same mistake so many times myself 😂 
```suggestion
                index = i
```

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataActivity.kt
```diff
@@ -51,44 +49,24 @@ class MetadataActivity : AppCompatActivity() , MetadataView, CaptureFlow {
...
+                break;
+            }
         }
+        if (index > -1)
+            spinner.setSelection(index)
```
eduardb, line 69: How would you actually feel rewriting this method in a more functional way like this?
```suggestion
        (0 until spinner.adapter.count)
            .firstOrNull { spinner.adapter.getItem(it) == value }
            ?.let { spinner.setSelection(it) }
```

app/src/main/java/net/aiscope/gdd_app/repository/SampleRepositorySharedPreference.kt
CeciliaHwang: Thanks for this suggestion! I'm making this change.

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataMapper.kt
CeciliaHwang: Thanks for the suggestion! I will make this change.

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataActivity.kt
CeciliaHwang: 🤦‍♀️Oops, thanks for catching that.

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataActivity.kt
CeciliaHwang: Much more succinct! Thanks for the suggestion. I am making this change.

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataMapper.kt
CeciliaHwang: That makes sense. Commit 1570e0fa414e42ef8f3280e75b6d8fb34fe28085 has a new MetadataMapperTest class with tests. Can you verify that this is the best way to write the tests?

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataPresenter.kt
CeciliaHwang: Passing the individual fields makes sense. If we change to do the mapping with MetadataMapper to be in MetadataPresenter, does it make sense to move the reverse mapping (using MetadataMapper) that is done during save to also be in the MetadataPresenter rather than MetadataActivity?

eduardb: Yes, I would think so :)

app/src/test/java/net/aiscope/gdd_app/presentation/MetadataMapperTest.kt
```diff
@@ -0,0 +1,103 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.Mock
+import org.mockito.junit.MockitoJUnitRunner
+
+@ExperimentalCoroutinesApi
+@RunWith(MockitoJUnitRunner::class)
+class MetadataMapperTest {
+    @get:Rule
+    var coroutinesTestRule = CoroutineTestRule()
```
eduardb, line 24: The `MetadataMapper` doesn't make use of coroutines, so we don't need this test rule.

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataPresenter.kt
```diff
@@ -34,7 +37,17 @@ class MetadataPresenter @Inject constructor(
...
         }
 
-        view.fillForm(ViewStateModel(sample.disease, sample.images.toList(), emptyList()))
+        val lastMetadata = repository.last()?.metadata
+        view.fillForm(
+            ViewStateModel(
+                sample.disease,
+                sample.images.toList(),
+                emptyList(),
+                smearType = lastMetadata?.let { MetadataMapper.getSmearTypeId(it.smearType) },
```
eduardb, line 46: This is correct as it is, but just as a personal preference, even if we have a singleton/object, I like to hide this kind of implementation detail by actually injecting the instance of it. Just something for you to consider if you want, I am not strongly holding to it.

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataMapper.kt
eduardb: They look good to me, I don't think that there's a better way due to those `Context.getString()` calls. The only thing I see is that test rule for coroutines, which is not used.

app/src/test/java/net/aiscope/gdd_app/presentation/MetadataMapperTest.kt
CeciliaHwang: Thanks. I removed this rule in changeset c0f5c9692349070bbd81db509eab0ed21928085e.

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataPresenter.kt
CeciliaHwang: Can you review changeset c0f5c9692349070bbd81db509eab0ed21928085e for this change? I wanted to confirm that you meant that I should pass in the instance of the SampleMetadata to the getSmearTypeId, etc. functions.

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataPresenter.kt
CeciliaHwang: Thanks, I made this change in changeset c0f5c9692349070bbd81db509eab0ed21928085e.

CeciliaHwang:
@eduardb I rebased from master before committing and pushing (and also without verifying first that the build was successful from master), so now I encounter detekt and unit test failures  from the master commits. Should I try to fix these issues, or expect that the committer will address them at some point and so I can ignore them?

CeciliaHwang:
> @eduardb I rebased from master before committing and pushing (and also without verifying first that the build was successful from master), so now I encounter detekt and unit test failures from the master commits. Should I try to fix these issues, or expect that the committer will address them at some point and so I can ignore them?

Please disregard. I just saw the slack thread on this, so I'll rebase again after it's resolved. 😀

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataPresenter.kt
eduardb: I was referring to the instance of `MetadataMapper` :P Sorry for not being clear. I can create a small commit to show you what I meant.

eduardb: https://github.com/eduardb/GDD-app/commit/02844d765e137fbc7271ec96eb4c53e959677feb
Feel free to use the same approach.

CeciliaHwang: Thanks for the code assist! I'm adding it exactly, except reverting back to passing in the values (smearType, species, stage) instead of the entire SampleMetadata.

app/src/main/java/net/aiscope/gdd_app/dagger/AppComponent.kt
```diff
@@ -4,6 +4,7 @@ import dagger.BindsInstance
 import dagger.Component
 import dagger.android.AndroidInjectionModule
 import net.aiscope.gdd_app.application.GddApplication
+import net.aiscope.gdd_app.ui.metadata.MetadataMapper
```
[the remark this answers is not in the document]
CeciliaHwang: Oops, thanks for catching that. I'm removing the reference.

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataMapper.kt
```diff
@@ -0,0 +1,74 @@
...
+import net.aiscope.gdd_app.R
+import net.aiscope.gdd_app.model.MalariaSpecies
+import net.aiscope.gdd_app.model.MalariaStage
+import net.aiscope.gdd_app.model.SampleMetadata
+import net.aiscope.gdd_app.model.SmearType
+
+object MetadataMapper {
+    fun getSmearType(smearTypeId: Int): SmearType {
+        return when (smearTypeId) {
+            R.id.metadata_blood_smear_thick -> SmearType.THICK
```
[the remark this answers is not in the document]
CeciliaHwang: 🤔

app/src/main/java/net/aiscope/gdd_app/repository/SampleRepositorySharedPreference.kt
```diff
@@ -47,4 +49,15 @@ class SampleRepositorySharedPreference @Inject constructor(
             gson.fromJson<SampleDto>(it).toDomain()
         }.toList()
     }
+
+    override suspend fun last(): Sample? {
+        val allStores = all()
+        val facility = healthFacilityRepository.load()
```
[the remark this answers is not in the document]
CeciliaHwang: I will remove the facility and microscopist fiter. Just curious, does each user store to a different database? That is, would pulling the last Sample potentially pull the Sample for another user?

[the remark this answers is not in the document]
eduardb: That could happen, yes. But realistically speaking, it's going to be very rare that multiple people use the same phone, so I wouldn't be too worried about this.

PR #25, opened 2020-04-22
[67] Display feedback once the user submits their sample

huanlui:
@daferpi 

Shows a Snackbar when the sample is saved successfully or an error occurs while saving. 

Below you can see the screenshots for both cases:

- Result for success:

image

- Result for error:

image

Some remarks:

- Snackbar functionality has been placed in a separate file so that new snackbars can be shown from other Views easily and uniformly .  
- A `Espresso` test for the happy path has been added in `MainActivityTest`. The error case has not been automatically tested because we don't know how to simulate an error when testing from `Espresso`.

Pairing with @MagdaSbant.


huanlui:
Failing but not able to see why:

image

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataActivity.kt
```diff
@@ -80,8 +82,35 @@ class MetadataActivity : AppCompatActivity() , MetadataView, CaptureFlow {
...
+        coroutineScope.launch {
+            try {
+                presenter.save(
+                    metadata_section_smear_type_radio_group.checkedRadioButtonId,
+                    metadata_species_spinner.selectedItem.toString(),
+                    metadata_stage_spinner.selectedItem.toString()
+                )
+                finishFlow()
+            }
+            catch(@Suppress("TooGenericExceptionCaught") error: Throwable) {
```
huanlui, line 95: Suppressed this rule as in `UploadWorker`

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataActivity.kt
```diff
@@ -80,8 +82,35 @@ class MetadataActivity : AppCompatActivity() , MetadataView, CaptureFlow {
...
-        goToHome()
+    private fun save() {
+        coroutineScope.launch {
+            try {
+                presenter.save(
+                    metadata_section_smear_type_radio_group.checkedRadioButtonId,
+                    metadata_species_spinner.selectedItem.toString(),
+                    metadata_stage_spinner.selectedItem.toString()
+                )
+                finishFlow()
```
daferpi, line 93: I don´t understand why you have move the method from the presenter to the activity, this is not the normal way to proceed, normally you need to execute the logic in the presenter and show the result in the activity. Also if you need to catch any error you need to do it in the presenter and have a method to success and a method to error that calls the activity.... i don´t understand to much the reason to do in this way, is possible to change?

app/src/main/java/net/aiscope/gdd_app/extensions/Snackbar.kt
```diff
@@ -0,0 +1,44 @@
...
+import androidx.annotation.StringRes
+import androidx.core.text.color
+import androidx.core.text.scale
+import com.google.android.material.snackbar.BaseTransientBottomBar
+import com.google.android.material.snackbar.Snackbar
+import net.aiscope.gdd_app.R
+
+const val SNACKBAR_TEXT_SCALE = 1.2f
+
+fun Snackbar.setActionOnDismissed(action: () -> Unit): Snackbar {
```
daferpi, line 17: What is the reason to create this extension methods? after review the code, maybe is a better idea build a custom component instead to use the android component?

app/src/main/java/net/aiscope/gdd_app/ui/CaptureFlow.kt
```diff
@@ -26,8 +26,9 @@ fun <T> T.attachCaptureFlowToolbar(toolbar: androidx.appcompat.widget.Toolbar)
     }
 }
 
-fun <T> T.goToHome() where T : AppCompatActivity, T : CaptureFlow {
+fun <T> T.goToHome(sampleSaved: Boolean = false) where T : AppCompatActivity, T : CaptureFlow {
```
daferpi, line 29: instead to pass a param is possible to have 2 functions with different behaviour? or is better this solution? is possible that in the future, you need to add more params?? also you can save the param in preferences and recover in the other activity, can be this a best solution?

app/src/main/java/net/aiscope/gdd_app/ui/main/MainActivity.kt
```diff
@@ -45,6 +43,14 @@ class MainActivity : AppCompatActivity(), SelectDiseaseView, LogoutFLow {
                 presenter.saveDisease(dropdown_select_disease.selectedItem.toString())
             }
         }
+
+        if(intent.hasExtra("SAMPLE_SAVED")) {
+            Snackbar.make(
```
daferpi, line 48: This is the reason that i says if it is better to have a custom component that wraps this logic... make sense?

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataActivity.kt
huanlui: Thanks, I'll have a look!

huanlui: Modified attending your suggestion

app/src/main/java/net/aiscope/gdd_app/extensions/Snackbar.kt
huanlui: Extension methods removed. A custom component has been created.

app/src/main/java/net/aiscope/gdd_app/ui/CaptureFlow.kt
huanlui: A separate function has been created instead of using a parameter.

Regarding the comment about saving the param in preferences, it sounds strange to me to save in preferences something that belongs to the flow.

app/src/main/java/net/aiscope/gdd_app/ui/main/MainActivity.kt
huanlui: A new component has been created as mentioned above.

PR #28, opened 2020-05-04
117 - Introduces color choice in Mask activity

raul-oliver-tw:


app/src/main/java/net/aiscope/gdd_app/ui/capture/CaptureImagePresenter.kt
```diff
@@ -19,7 +19,7 @@ class CaptureImagePresenter(
                 val sample = repository.current().addImage(file)
                 repository.store(sample)
 
-                view.goToMask(file.absolutePath, nextMaskName)
+                view.goToMask(sample.disease!!, file.absolutePath, sample.nextMaskName())
```
daferpi, line 22: unwrap optionals without validation is dangerous and can crash the application, remember that we are working in kotlin but behind android works with java and java not manage nulls

raul-oliver-tw: What do you suggest in cases like this? I considered adding some validation in some cases like this one, but the outcome would, as I understand, the same: That's a result of some incorrectly implemented flow and the user may not do anything about it, so what would be the difference between the app crashing and the user getting a message about something he/she does not have any control? What are your thoughts? Thanks!!

PR #32, opened 2020-05-20
[104] Adds SamplePreparation and MicroscopeQuality screens

raul-oliver-tw:


app/src/main/java/net/aiscope/gdd_app/network/ModelDtoExt.kt
```diff
@@ -0,0 +1,39 @@
...
+import net.aiscope.gdd_app.repository.SampleMetadataDto
+import net.aiscope.gdd_app.repository.SamplePreparationDto
+
+fun Sample.toDto() = SampleDto(
+    id = id,
+    healthFacility = healthFacility,
+    microscopist = microscopist,
+    disease = disease ?: "",
+    preparation = preparation?.toDto(),
+    microscopeQuality = microscopeQuality?.toDto(),
```
jbors, line 18: Might be a good idea to add a test to see if the selected preparation & microscope show up in the JSON that is saved with the sample in the end

app/src/main/res/values/strings.xml
```diff
@@ -6,18 +6,43 @@
...
+    <string name="sample_preparation_water_type_title">Water</string>
+    <string name="sample_preparation_giemsa_title">Uses Giemsa</string>
+    <string name="sample_preparation_giemsa_fp_title">Giemsa filtering or precipitating</string>
+    <string name="sample_preparation_pbs_title">Uses PBS</string>
+    <string name="sample_preparation_alcohol_title">Uses alcohol to fix the sample</string>
+    <string name="sample_preparation_slides_reuse_title">Reuses slides</string>
+    <string name="sample_preparation_continue">continue</string>
+    <string name="sample_preparation_snackbar_error">An error occurred while saving your sample preparation data. Please try again.</string>
+    <string name="sample_preparation_snackbar_retry">TRY AGAIN</string>
+    <string name="water_type_bottled">bottled</string>
```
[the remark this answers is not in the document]
raul-oliver-tw: I was not sure about it, but making it consistent with the smear types sound good! Done, Thanks!

app/src/main/java/net/aiscope/gdd_app/network/ModelDtoExt.kt
raul-oliver-tw: Definitely! Done, thanks!

app/src/main/res/values/strings.xml
```diff
@@ -52,16 +76,13 @@
...
-    <string name="mask_activity_undo">Undo</string>
-    <string name="mask_activity_redo">Redo</string>
-    <string name="mask_activity_stages">Select stage</string>
-    <string name="mask_activity_stages_dialog_title"><font color="#ffffffff">Select a stage</font></string>
-    <string name="button_cancel">Cancel</string>
-    <string name="button_ok">OK</string>
+    <string name="microscope_quality_damaged_title">Microscope damaged/malfunctioning</string>
+    <string name="microscope_quality_magnification_title">Magnification</string>
+    <string name="microscope_quality_magnification_edit_text_hint">Enter a number between 0 and 2000</string>
+    <string name="mandatory_field_error">mandatory field</string>
```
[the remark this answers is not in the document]
raul-oliver-tw: Done, thanks!!

raul-oliver-tw:
> Couple small remarks.
> 
> Also, it seems that we get the 'Sample preparation' and 'Microscope quality' screens before _every_ sample. I was under the impression we would do this once & then apply it to all samples. Did the scope change for V1 maybe?

Thanks for your comments! Yes, it was agreed that we'd do it this way for v1 - but I still intend to push a little to take a step further and implement this change you're mentioning early on. :-)

app/src/main/java/net/aiscope/gdd_app/dagger/MicroscopeQualityModule.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    @Binds
+    @PerActivity
+    internal abstract fun provideView(view: MicroscopeQualityActivity): MicroscopeQualityView
+
+    @Binds
+    @PerActivity
+    internal abstract fun activity(activity: MicroscopeQualityActivity): Activity
+
+    @Module
```
jbors, line 22: These `@module` and `@jvmstatic` annotations have recently been removed for other modules. See https://github.com/theaiscope/GDD-app/commit/714614af91441f36e57b73f5e1c96646d0e5bb3f

I'm assuming we also won't need them here?

app/src/main/java/net/aiscope/gdd_app/dagger/SamplePreparationModule.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    @Binds
+    @PerActivity
+    internal abstract fun provideView(view: SamplePreparationActivity): SamplePreparationView
+
+    @Binds
+    @PerActivity
+    internal abstract fun activity(activity: SamplePreparationActivity): Activity
+
+    @Module
```
jbors, line 22: Same here

app/src/main/java/net/aiscope/gdd_app/extensions/SpinnerExt.kt
```diff
@@ -0,0 +1,9 @@
+package net.aiscope.gdd_app.extensions
+
+import android.widget.Spinner
+
+fun Spinner.select(value: String?) {
```
eduardb, line 5: Nice :)

app/src/main/java/net/aiscope/gdd_app/ui/main/MainActivity.kt
```diff
@@ -70,20 +71,11 @@ class MainActivity : AppCompatActivity(), SelectDiseaseView, LogoutFLow {
...
 
-    override fun captureImage(nextImageName: String) {
-        val intent = Intent(this, CaptureImageActivity::class.java)
-        intent.putExtra(CaptureImageActivity.EXTRA_IMAGE_NAME, nextImageName)
+    override fun goToSamplePreparation() {
+        val intent = Intent(this, SamplePreparationActivity::class.java)
         this.startActivity(intent)
     }
 
-    override fun showSuccessToast() {
```
eduardb, line 79: 💯

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataActivity.kt
```diff
@@ -3,19 +3,22 @@ package net.aiscope.gdd_app.ui.metadata
...
 import android.view.View
-import android.widget.AbsSpinner
 import android.widget.Toast
 import androidx.appcompat.app.AppCompatActivity
 import androidx.recyclerview.widget.LinearLayoutManager
 import com.google.android.material.snackbar.Snackbar
 import dagger.android.AndroidInjection
-import kotlinx.android.synthetic.main.activity_metadata.*
-import kotlinx.android.synthetic.main.toolbar.*
+import kotlinx.android.synthetic.main.activity_metadata.metadata_blood_sample_images
```
eduardb, line 11: Not sure about these synthetic imports, do you feel that it's better without the star?
Btw, maybe you want to have a look on the new shiny thing, [View Binding](https://developer.android.com/topic/libraries/view-binding).

app/src/main/res/layout/activity_microscope_quality.xml
```diff
@@ -0,0 +1,70 @@
...
+    android:layout_height="match_parent"
+    android:orientation="vertical"
+    tools:context=".ui.microscope_quality.MicroscopeQualityActivity">
+
+    <include
+        layout="@layout/toolbar"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent" />
+
+    <androidx.constraintlayout.widget.ConstraintLayout
```
eduardb, line 16: Why is a ConstraintLayout needed in another ConstraintLayout? 🙈

app/src/main/res/layout/activity_sample_preparation.xml
```diff
@@ -0,0 +1,99 @@
...
+    android:layout_height="match_parent"
+    android:orientation="vertical"
+    tools:context=".ui.sample_preparation.SamplePreparationActivity">
+
+    <include
+        layout="@layout/toolbar"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent" />
+
+    <androidx.constraintlayout.widget.ConstraintLayout
```
eduardb, line 16: Same question here.

app/src/main/res/values/strings.xml
```diff
@@ -6,19 +6,44 @@
     <string name="error_message_field_empty">Field cannot be empty</string>
     <string name="error_message_logout_failure">Error when during logout process</string>
     <string name="confirmation_message_saved">Saved</string>
+    <string name="button_cancel">Cancel</string>
+    <string name="button_ok">OK</string>
```
eduardb, line 10: IIRC, "OK" and "Cancel" are already defined in `android.R.string.*`, maybe you want to use those, they are translated too.

app/src/main/res/values/styles.xml
```diff
@@ -19,18 +19,32 @@
...
         <item name="windowNoTitle">true</item>
     </style>
 
+    <style name="AppTheme.NoActionBar.SamplePreparationActivity">
+        <item name="switchStyle">@style/Switch</item>
+    </style>
+
+    <style name="AppTheme.NoActionBar.MicroscopeQualityActivity">
+        <item name="switchStyle">@style/Switch</item>
+    </style>
```
eduardb, line 36: Hmm, why having two new identical styles for the new activities? Couldn't the `switchStyle` be added to the parent style/theme?

app/src/main/res/values/styles.xml
```diff
@@ -1,5 +1,5 @@
 <resources>
-    <style name="AppTheme" parent="Theme.AppCompat.Light">
+    <style name="AppTheme" parent="Theme.AppCompat">
```
eduardb, line 2: Doesn't this break anything? 😓

app/src/main/java/net/aiscope/gdd_app/model/MalariaSpecies.kt
```diff
@@ -0,0 +1,10 @@
+package net.aiscope.gdd_app.model
+
+@Suppress("MagicNumber") // these are IDs
+enum class MalariaSpecies(val id: Int) {
```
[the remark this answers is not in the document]
jbors: Like having one 'Model.kt' file with all these classes you mean?

app/src/main/java/net/aiscope/gdd_app/dagger/MicroscopeQualityModule.kt
raul-oliver-tw: Great catch! Thanks!

app/src/main/java/net/aiscope/gdd_app/dagger/SamplePreparationModule.kt
raul-oliver-tw: Thanks!

app/src/main/java/net/aiscope/gdd_app/dagger/MicroscopeQualityModule.kt
```diff
@@ -0,0 +1,29 @@
...
+@Module
+abstract class MicroscopeQualityModule {
+
+    @Binds
+    @PerActivity
+    internal abstract fun provideView(view: MicroscopeQualityActivity): MicroscopeQualityView
+
+    @Binds
+    @PerActivity
+    internal abstract fun activity(activity: MicroscopeQualityActivity): Activity
```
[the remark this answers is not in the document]
raul-oliver-tw: Oops. Thanks!

app/src/main/java/net/aiscope/gdd_app/dagger/BuildersModule.kt
```diff
@@ -27,8 +29,14 @@ abstract class BuildersModule {
 
     @Binds
     @IntoMap
-    @ClassKey(MetadataActivity::class)
-    abstract fun bindMetadataActivityInjectorFactory(factory: MetadataSubComponents.Factory):
+    @ClassKey(SamplePreparationActivity::class)
+    abstract fun bindSamplePreparationActivityInjectorFactory(factory: SamplePreparationSubComponents.Factory):
```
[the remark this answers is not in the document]
raul-oliver-tw: Great, thanks!!

app/src/main/res/layout/activity_microscope_quality.xml
raul-oliver-tw: XD Just a matter of organisation, with some subtle consequences, like the inner ConstraintLayout padding, which replaced the padding that would be added to every element within it... Is that as bad as an anti-pattern? How could we tackle that attribute's repetition, if not with this approach? Thanks!

app/src/main/res/layout/activity_sample_preparation.xml
raul-oliver-tw: Resolving conversation - will do whatever we agree on both activities. Thanks!!

app/src/main/res/values/strings.xml
raul-oliver-tw: Hmmmm... Not my strings - I just reorganised them by "scopes"... Could not find where these were used, though, so I'm removing them. Thanks!!

app/src/main/res/values/styles.xml
raul-oliver-tw: Yep, definitely... I just tried to keep the styles as flexible and independent as possible, as I'm not sure where we're going regarding styling. I'd like, in fact, to not need to create any major styling customisation (which I think is the case here). Anyway, do you get my point? What do you think would be healthier regarding this?

app/src/main/res/values/styles.xml
raul-oliver-tw: Errrr... This was accidental actually - every time I touch styling in this project I try this theme as a parent to try to understand why it was not used in the first place, but could not get to a conclusion yet... XD Do you have any info on that?

app/src/main/java/net/aiscope/gdd_app/model/MalariaSpecies.kt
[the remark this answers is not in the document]
raul-oliver-tw: I'd rather keeping each class in its file, but that's probably a result of my familiarity with Java... XD I think @jbors comment makes total sense, so I'm putting it all together again, but renaming the file to "Model.kt"...

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataActivity.kt
raul-oliver-tw: Yes, normally I'd go without any star import, but I agree the synthetic ones are a different case. I had removed their aggregation on the projects config, but then I reverted that (kept the no-grouping for java.util, though). I just pushed a commit with imports reorganisation for the activities. Thanks for the View Binding reference - I think I like it, have to try! :-D

app/src/main/java/net/aiscope/gdd_app/model/MalariaSpecies.kt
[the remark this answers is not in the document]
eduardb: I file called `Model.kt` in the `.model` package? I think that we can do better :P 
How about just `Sample.kt`? (since `Sample` is the main domain model`

app/src/main/res/layout/activity_microscope_quality.xml
eduardb: I see. I think that there are two possible solutions: 
- One outer ConstraintLayout with Guidelines for the padding.
or
- Use a LinearLayout with vertical orientation instead of the inner ConstraintLayout.

Probably the latter is easier/faster to achieve.

app/src/main/res/values/styles.xml
eduardb: > flexibile

Yes, please.

> independent

I think that this is counter-productive when it comes to styles, as it leads to a lot of accidental variation throughout the app. But normally, you want consistency in the design of your app. 

So I would say to consolidate the main theme of the app and use just that, instead of creating additional themes. Anytime something needs to have a different style, we can do it per view or activity/view group level easily.

app/src/main/res/values/styles.xml
eduardb: I think that some activities have their own `Toolbar` in their layout, while some don't (the Firebase UI login one actually), so that's why the `.NoActionBar` one is used. Did I explain myself well? :D

app/src/main/java/net/aiscope/gdd_app/model/MalariaSpecies.kt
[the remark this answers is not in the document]
raul-oliver-tw: SampleModels.kt? I agree Models.kt in the models package is redundant, but hey: We're using the Kotlin file with the same purpose of a package, aren't we? Sample.kt sounds off to me as well, as there is not only the Sample model in there... So, how about SampleModels.kt? Do you like the idea?

app/src/main/res/layout/activity_microscope_quality.xml
raul-oliver-tw: The first one is not really an option, as the padding would be applied to all elements, including the toolbar and the action bar (I tried that when going through this XD).
I had thought about the second one, but what is the advantage on that? (I'm asking, not challenging) I'd expect the LinearLayout to be better performant than the ConstraintLayout, but what I read is that ConstraintLayout should be used whenever possible (could not get to a conclusion on the reason - maybe to try to get rid of nested layouts like this one?). Thanks!!

app/src/main/res/values/styles.xml
raul-oliver-tw: Sure, I got what you mean, makes sense. But I actually referred to the usage of ".Light" instead of the dark theme as a parent of the main AppTheme... Any reason for that that you know?

app/src/main/res/layout/activity_microscope_quality.xml
raul-oliver-tw: Is there any way to make the LinearLayout fill the whole screen vertically without needing to use a spacer? (ideally the inner ConstraintLayout should fill its parent)

app/src/main/java/net/aiscope/gdd_app/ui/microscope_quality/MicroscopeQualityPresenter.kt
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    suspend fun save(viewModel: MicroscopeQualityViewStateModel) {
+        val model = mapper.convert(viewModel)
+        try {
+            val sample = repository.current()
+                .copy(microscopeQuality = model)
+            repository.store(sample)
+            view.goToCaptureImage(sample.nextImageName())
+        } catch(@Suppress("TooGenericExceptionCaught") error: Throwable) {
```
eduardb, line 27: What errors can happen here? For me, the `TooGenericExceptionCaught` is a smell of something bad that might happen which might just cause the application to misbehave instead of letting it crash, so we can get a Crashlytics report with the issue.

app/src/main/res/values/strings.xml
```diff
@@ -6,19 +6,42 @@
...
     <string name="image_could_not_be_taken">Image could not be taken</string>
-
     <!-- Mask -->
     <string name="mask_activity_title">Highlight</string>
     <string name="button_done_text">Done</string>
-
+    <string name="mask_activity_undo">Undo</string>
+    <string name="mask_activity_redo">Redo</string>
+    <string name="mask_activity_stages">Select stage</string>
+    <string name="mask_activity_stages_dialog_title"><font color="#ffffffff">Select a stage</font></string>
```
eduardb, line 44: I think that this string is not used, no? Could it be removed then?

app/src/main/res/values/styles.xml
raul-oliver-tw: Done, thanks!

app/src/main/res/layout/activity_microscope_quality.xml
eduardb: > The first one is not really an option, as the padding would be applied to all elements

With a [`Guideline`](https://constraintlayout.com/basics/guidelines.html), you can choose what elements "have" padding, and which not. 

> ConstraintLayout should be used whenever possible

The idea of the `ConstraintLayout` is to have a flat hierarchy, so yes, it should be used whenever possible, but having nested ones defeats its purpose. And yes, the ultimate price that you are paying is performance.

> Is there any way to make the LinearLayout fill the whole screen vertically without needing to use a spacer? (ideally the inner ConstraintLayout should fill its parent)

Yes, let the `LinearLayout` have `android:layout_height="match_parent"`, and then on the widget that you want to take up all the available space put the following:
```xml
        android:layout_height="0dp"
        android:layout_weight="1"
```

app/src/main/res/values/styles.xml
eduardb: Oh, sorry for the misunderstanding. I am not sure about that, probably just the color scheme used in the app better matches that?

app/src/main/res/layout/activity_microscope_quality.xml
```diff
@@ -0,0 +1,70 @@
...
+
+        <TextView
+            android:id="@+id/microscope_quality_magnification_error"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:text="@string/microscope_quality_magnification_error"
+            android:visibility="gone"
+            app:layout_constraintStart_toStartOf="@id/microscope_quality_magnification_edit_text"
+            app:layout_constraintTop_toBottomOf="@id/microscope_quality_magnification_edit_text"
+            android:textColor="@color/colorError" />
```
eduardb, line 58: These three widgets can be changed for a [`TextInputLayout`](https://material.io/develop/android/components/text-fields/) to achieve the same functionality in a more interactive and Material-like way. What do you think, would it be worth using that instead?

app/src/androidTest/java/net/aiscope/gdd_app/model/SampleTest.kt
```diff
@@ -15,16 +17,47 @@ class SampleTest {
...
+        )
+
+        val expectedMicroscopeQualityDto = MicroscopeQualityDto(
+            true,
+            1240
+        )
+
+        assertEquals("_id", dto.id)
+        assertEquals("facility", dto.healthFacility)
+        assertEquals("microscopist", dto.microscopist)
```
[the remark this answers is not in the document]
raul-oliver-tw: Done, thanks!

app/src/main/java/net/aiscope/gdd_app/dagger/SamplePreparationModule.kt
```diff
@@ -0,0 +1,27 @@
...
+@Module
+abstract class SamplePreparationModule {
+
+    @Binds
+    @PerActivity
+    internal abstract fun provideView(view: SamplePreparationActivity): SamplePreparationView
+
+    @Binds
+    @PerActivity
+    internal abstract fun activity(activity: SamplePreparationActivity): Activity
```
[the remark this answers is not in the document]
raul-oliver-tw: Indeed - removed. Thanks!

app/src/main/java/net/aiscope/gdd_app/model/Model.kt
```diff
@@ -2,40 +2,24 @@ package net.aiscope.gdd_app.model
...
-
-@Suppress("MagicNumber") // these are IDs
-enum class MalariaSpecies(val id: Int) {
-    P_FALCIPARUM(1),
-    P_VIVAX(2),
-    P_OVALE(3),
-    P_MALARIAE(4),
-    P_KNOWLESI(5)
-}
+data class Disease(val name: String)
```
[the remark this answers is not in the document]
raul-oliver-tw: No idea... Just removed it - we'll figure what it was used for if something crashes... XD Thanks!

app/src/main/java/net/aiscope/gdd_app/ui/microscope_quality/MicroscopeQualityActivity.kt
```diff
@@ -0,0 +1,111 @@
...
+import net.aiscope.gdd_app.ui.snackbar.CustomSnackbarAction
+import javax.inject.Inject
+
+class MicroscopeQualityActivity : AppCompatActivity(), MicroscopeQualityView, CaptureFlow {
+
+    companion object {
+        private const val MAGNIFICATION_MIN = 0
+        private const val MAGNIFICATION_MAX = 2000
+        private const val DEFAULT_DAMAGED = false
+        private const val DEFAULT_MAGNIFICATION = 1000
```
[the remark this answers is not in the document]
raul-oliver-tw: I initially took that approach, but then I came to the conclusion that these are not the model's default values as such - they are just what we expect to be the most used values, so we're pre-populating the fields with those values to facilitate the UX, and that's a part of the presentation layer. Does that make sense?

app/src/main/java/net/aiscope/gdd_app/ui/microscope_quality/MicroscopeQualityActivity.kt
```diff
@@ -0,0 +1,111 @@
...
+            magnification = DEFAULT_MAGNIFICATION
+        } else {
+            isDamaged = model.isDamaged
+            magnification = model.magnification
+        }
+        microscope_quality_damaged_switch.isChecked = isDamaged
+        microscope_quality_magnification_edit_text.setText(magnification.toString())
+    }
+
+    private fun validateForm(): Boolean {
```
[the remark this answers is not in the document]
raul-oliver-tw: The way I understood it, the presenter should handle (business-logical) data, while the view should handle the components... Validating is handling data, ok, but it's data contained in the components... XD These validations are way too simple, and creating something to conform these "rules" would cost us much more effort than the actual validation... So, I ask you back: Do you think it's worth the effort?

app/src/main/java/net/aiscope/gdd_app/ui/microscope_quality/MicroscopeQualityPresenter.kt
raul-oliver-tw: I agree - I just made it consistent with what's done in the Metadata screen... We should maybe raise a ticket to review these points?

app/src/main/res/values/strings.xml
raul-oliver-tw: Thanks for noticing! Removed.

app/src/main/res/layout/activity_microscope_quality.xml
raul-oliver-tw: Great tips there!! Thanks!

app/src/main/java/net/aiscope/gdd_app/model/MalariaSpecies.kt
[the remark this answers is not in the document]
raul-oliver-tw: Agreed with @eduardb having SampleModels.kt by dm. Thanks!!

app/src/main/res/values/styles.xml
raul-oliver-tw: @eduardb told my through dm that it's probably due to FirebaseUI styling. Thanks, @eduardb!

app/src/main/res/layout/activity_microscope_quality.xml
raul-oliver-tw: Great call, thanks!

app/src/main/java/net/aiscope/gdd_app/ui/microscope_quality/MicroscopeQualityActivity.kt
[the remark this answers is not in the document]
raul-oliver-tw: Gave it a little refactoring - it seems a bit more coupled, but we already had the same coupling with the object coming from the repo, and now the code looks more organised. Do you like it? Thanks!

app/src/main/java/net/aiscope/gdd_app/ui/sample_preparation/SamplePreparationActivity.kt
```diff
@@ -0,0 +1,110 @@
...
+            usesPbs = true,
+            usesAlcohol = true,
+            reusesSlides = false
+        )
+    }
+
+    @Inject
+    lateinit var presenter: SamplePreparationPresenter
+
+    private val parentJob = Job()
```
eduardb, line 40: `parentJob.cancel()` should be called in `onDestroy()` to cancel any running jobs. But now in Android Lifecycle Runtime KTX there is `LifecycleOwner.lifecycleScope`, so you could use `lifecycleScope` instead of `coroutineScope` and get rid of the fluff :)

app/src/main/java/net/aiscope/gdd_app/ui/sample_preparation/SamplePreparationActivity.kt
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    override fun fillForm(model: SamplePreparationViewStateModel?) {
+        val formData = model ?: defaultFormData
+        sample_preparation_water_type_spinner.select(formData.waterType)
+        sample_preparation_giemsa_switch.isChecked = formData.usesGiemsa
+        sample_preparation_giemsa_fp_switch.isChecked = formData.giemsaFP
+        sample_preparation_pbs_switch.isChecked = formData.usesPbs
+        sample_preparation_alcohol_switch.isChecked = formData.usesAlcohol
+        sample_preparation_slides_reuse_switch.isChecked = formData.reusesSlides
```
eduardb, line 67: For a bit less repetition you could do something like: 
```suggestion
        with(formData) {
            sample_preparation_water_type_spinner.select(waterType)
            sample_preparation_giemsa_switch.isChecked = usesGiemsa
            sample_preparation_giemsa_fp_switch.isChecked = giemsaFP
            sample_preparation_pbs_switch.isChecked = usesPbs
            sample_preparation_alcohol_switch.isChecked = usesAlcohol
            sample_preparation_slides_reuse_switch.isChecked = reusesSlides
        }
```

app/src/main/java/net/aiscope/gdd_app/ui/sample_preparation/SamplePreparationActivity.kt
raul-oliver-tw: Great, thanks! :-)

app/src/main/java/net/aiscope/gdd_app/ui/microscope_quality/MicroscopeQualityActivity.kt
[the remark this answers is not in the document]
raul-oliver-tw: @eduardb approved the change in another comment - resolving this one.

app/src/main/java/net/aiscope/gdd_app/ui/microscope_quality/MicroscopeQualityActivity.kt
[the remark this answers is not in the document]
raul-oliver-tw: Thanks!

app/src/main/java/net/aiscope/gdd_app/ui/microscope_quality/MicroscopeQualityPresenter.kt
raul-oliver-tw: Created Trello card https://trello.com/c/HVEezUXD. Thanks!

PR #35, opened 2020-05-21
[140] Adds app version to main screen

raul-oliver-tw:


app/src/main/java/net/aiscope/gdd_app/ui/main/MainActivity.kt
```diff
@@ -36,6 +38,9 @@ class MainActivity : AppCompatActivity(), SelectDiseaseView, LogoutFLow {
         setContentView(R.layout.activity_main)
         setSupportActionBar(toolbar)
 
+        @SuppressLint("SetTextI18n")
+        main_version_text.text = "v${BuildConfig.VERSION_NAME}"
```
eduardb, line 42: What do you think about using a formatted string resource instead of hardcoding the string literal here?
See https://developer.android.com/guide/topics/resources/string-resource.html#formatting-strings

raul-oliver-tw: I didn't know about that possibility - much better! Thanks!!

PR #38, opened 2020-05-26
[122] Show Privacy Policy after login

eduardb:
This adds a privacy policy that needs to be agreed to before using the app. 

The policy is loaded into a WebView by using a static HTML that makes a `fetch()` call in JavaScript to get the up-to-date text from the AI Scope website. A bridge is used to notify the app when the policy has been actually loaded.

|Screenshot|
|----------|
||


app/src/test/java/net/aiscope/gdd_app/presentation/LoginPresenterTest.kt
```diff
@@ -0,0 +1,47 @@
...
+import net.aiscope.gdd_app.repository.MicroscopistRepository
+import net.aiscope.gdd_app.ui.login.LoginPresenter
+import net.aiscope.gdd_app.ui.login.LoginView
+import org.junit.Test
+
+class LoginPresenterTest {
+
+    companion object {
+        val microscopistThatAgreedPrivacyPolicy = Microscopist("1", true)
+        val microscopistThatDidntAgreedPrivacyPolicy = Microscopist("1", false)
```
eduardb, line 18:
```suggestion
        val microscopistThatHasntAgreedPrivacyPolicy = Microscopist("1", false)
```
Ups, grammar

app/src/main/java/net/aiscope/gdd_app/ui/policy/PrivacyPolicyActivity.kt
```diff
@@ -0,0 +1,118 @@
...
+    @Inject
+    lateinit var presenter: PrivacyPolicyPresenter
+
+    private lateinit var binding: ActivityPrivacyPolicyBinding
+
+    @SuppressLint("SetJavaScriptEnabled")
+    override fun onCreate(savedInstanceState: Bundle?) {
+        AndroidInjection.inject(this)
+        super.onCreate(savedInstanceState)
+        binding = ActivityPrivacyPolicyBinding.inflate(layoutInflater)
```
raul-oliver-tw, line 33: Nice! Is there a way to prevent the synthetics creation?

app/src/main/java/net/aiscope/gdd_app/ui/policy/PrivacyPolicyActivity.kt
```diff
@@ -0,0 +1,118 @@
...
+
+    override fun goToMain() {
+        finish()
+        startActivity(Intent(this, MainActivity::class.java))
+    }
+
+    private fun setUpWebView() {
+        WebView.setWebContentsDebuggingEnabled(true)
+        with(binding.webView) {
+            settings.javaScriptEnabled = true
```
raul-oliver-tw, line 63: No way to set this through the XML, right? (I googled for it but couldn't find anything)

app/src/main/java/net/aiscope/gdd_app/ui/policy/PrivacyPolicyActivity.kt
```diff
@@ -0,0 +1,118 @@
...
+    override fun showError(errorMessage: String?) = with(binding) {
+        error.isVisible = true
+        error.text = getString(R.string.privacy_policy_error_message, errorMessage)
+        agreeButton.isVisible = false
+        cancelButton.isVisible = false
+        progressBar.isVisible = false
+    }
+}
+
+class Callback(val presenter: PrivacyPolicyPresenter) {
```
raul-oliver-tw, line 102: I know this in obvious in the context, but can we rename this to something more explicit? PrivacyPolicyLoadCallback maybe?

app/src/main/java/net/aiscope/gdd_app/ui/policy/PrivacyPolicyActivity.kt
eduardb: You mean the kotlinx ones? Yes, by removing the `kotlin-android-extensions` Gradle plugin, but we would have to migrate to View Binding everywhere first :))

app/src/main/java/net/aiscope/gdd_app/ui/policy/PrivacyPolicyActivity.kt
eduardb: Nope, unfortunately :(

app/src/main/java/net/aiscope/gdd_app/ui/policy/PrivacyPolicyActivity.kt
raul-oliver-tw: Nice! Thanks! :-)

app/src/main/java/net/aiscope/gdd_app/ui/policy/PrivacyPolicyActivity.kt
eduardb: Good point, done in 1d0fbe1

eduardb:
You noticed really well, but this is a trade-off we decided to make for having this sooner. See this thread on Slack: https://aiscope.slack.com/archives/CMUEVLWRG/p1589892033002900

app/build.gradle
```diff
@@ -86,6 +90,7 @@ dependencies {
     implementation 'com.github.salomonbrys.kotson:kotson:2.5.0'
     implementation 'androidx.work:work-runtime-ktx:2.3.4'
     implementation 'androidx.core:core-ktx:1.2.0'
+    implementation 'androidx.activity:activity-ktx:1.2.0-alpha04'
```
raul-oliver-tw, line 93: I had to copy this to my PR because of you suggestion on lifecycleScope and Android Studio told me this is not the latest available version - I'm assuming you're aware and the choice was intentional.

eduardb: Oh, I wasn't aware that this has `lifecycle-runtime-ktx` as a transitive dependency, haha. 
I added this dependency to make use of the new `ActivityResultContracts` in the `LoginActivity`.

eduardb: Now I understood what you meant, I will be updating to a newer version :D

raul-oliver-tw: Thanks!! :-)

PR #43, opened 2020-06-08
[138] Save timestamp

jbors:
Save timestamp of latest user save to firebase. We chose latest user save to allow for editing of samples later. 

--UPDATE: Now with immutable domain object and saving the createdOn date as well

### DIFF
--- a/app/src/main/java/net/aiscope/gdd_app/model/SampleModels.kt
+++ b/app/src/main/java/net/aiscope/gdd_app/model/SampleModels.kt
@@ -16,7 +16,8 @@ data class Sample(
     val masks: LinkedHashSet<File> = linkedSetOf(),
     val metadata: SampleMetadata = SampleMetadata(),
     val status: SampleStatus = SampleStatus.Incomplete,
-    val createdOn: Calendar? = null
+    val createdOn: Calendar? = null,
+    var lastModified: Calendar? = null
 ) {
     fun addImage(path: File) = copy(images = images + path)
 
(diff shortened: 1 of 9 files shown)


app/src/main/java/net/aiscope/gdd_app/model/SampleModels.kt
```diff
@@ -16,7 +16,8 @@ data class Sample(
     val masks: LinkedHashSet<File> = linkedSetOf(),
     val metadata: SampleMetadata = SampleMetadata(),
     val status: SampleStatus = SampleStatus.Incomplete,
-    val createdOn: Calendar? = null
+    val createdOn: Calendar? = null,
+    var lastModified: Calendar? = null
```
eduardb, line 20: Are there any specific reasons why you went for a mutable property?

jbors: @eduardb Well we set it on 'store' rather than object creation. Also, there is a wish to be able to come back & make changes to this later, so conceptually I think it makes sense to have this as a mutable prop. On any update the 'lastModified' should change.

eduardb: I mean, I agree, but the class has been originally made immutable on purpose, as this is good practice in a multi-threaded environment :) Also, it's possible to achieve the requirement by using the `copy()` method, and then return the newly updated object to the caller.
Again, I am not strong on this, feel free to merge as it is.

jbors: @eduardb  Hmmm yeah you may have a point. I also noticed another thing that could be improved. Will have another pass over this. Thanks!

PR #44, opened 2020-06-14
Smoother drawing + points

eduardb:
The changes in this PR allows single points to be drawn.

Also, it makes drawing smoother by registering all moves, including the ones smaller than `ViewConfiguration.scaledTouchSlop` (this should be used only for scrolling, I mistakenly added it in the past for drawing).

| Before | After |
|--------|-------|
| before | after |

### DIFF
--- a/app/src/main/java/net/aiscope/gdd_app/ui/mask/customview/MaskLayer.kt
+++ b/app/src/main/java/net/aiscope/gdd_app/ui/mask/customview/MaskLayer.kt
@@ -204,15 +204,7 @@ class MaskLayer(private val imageMatrix: Matrix) {
     }
 
     fun drawMove(x: Float, y: Float) {
-        currentPath?.run {
-            val (latestX, latestY) = latestPoint
-            val dX = abs(x - latestX)
-            val dY = abs(y - latestY)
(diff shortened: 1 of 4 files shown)


app/src/main/java/net/aiscope/gdd_app/ui/mask/customview/MaskLayer.kt
```diff
@@ -204,22 +204,12 @@ class MaskLayer(
...
-        }
+        currentPath?.quadTo(x, y)
     }
 
     fun drawEnd() {
         currentPath?.run {
-            val visibleChange: Boolean =
-                if (isCurrentModeDraw()) hasMultiplePoints()
-                else !latestChangeBitmap.sameAs(currentStateBitmap)
+            val visibleChange: Boolean = !latestChangeBitmap.sameAs(currentStateBitmap)
```
jbors, line 212: Since we only use `visibleChange` in the if on the next line, I would prefer to simply have it in the if statement. Also, since we dropped the 'multiplepoints' requirement, I don't think `visible` is a useful name here

eduardb: Done in 4dc9120.

PR #48, opened 2020-07-29
[144] Add scrolling for landscape sample preparation and metadata.

anna-ruizferrer:
Add landscape view of main screen.


app/src/main/res/layout-land/activity_main.xml
```diff
@@ -0,0 +1,83 @@
...
+        android:theme="@style/ThemeOverlay.AppCompat.ActionBar"
+        app:contentInsetStart="40dp"
+        app:layout_constraintTop_toTopOf="parent"
+        tools:title="AI Scope" />
+
+    <ImageView
+        android:id="@+id/icon"
+        android:layout_width="match_parent"
+        android:layout_height="100dp"
+        android:layout_marginTop="20dp"
```
eduardb, line 23: I see that only some dimensions have changed from `layout/activity_main.xml`. In that case, you shouldn't create a new layout file for landscape, but you could just extract those dimensions into resources, and have landscape-specific dimensions in `values-land/dimens.xml`

PR #49, opened 2020-07-29
Migrate from kotlin-android-extension synthetic views to View Binding

eduardb:
This improves the naming of the view references in the code, and potentially having some performance benefits too. The drawback is that a `binding` object needs to be instantiated and used everywhere, but this seems to be the preferred way of referencing views on Android going forward.

Reference: https://developer.android.com/topic/libraries/view-binding

(Upgrade some Gradle dependencies along the way too)


app/build.gradle
```diff
@@ -33,7 +33,9 @@ android {
     buildFeatures {
         viewBinding = true
     }
-
+    androidExtensions {
```
eduardb, line 36: This will only let the `parcelize` feature active from the `kotlin-android-extensions` plugin, basically deactivating the synthetic views one. We might use this feature to create `Parcelable` classes just with an annotation :D

jbors: I mean, it's cool. But if we're gonna introduce the dependency shouldn't we use `@Parcelize` on our `Parcelable` classes while we're at it?

app/src/main/java/net/aiscope/gdd_app/ui/mask/MaskActivity.kt
```diff
@@ -107,8 +112,6 @@ class MaskActivity : AppCompatActivity(), MaskView, CaptureFlow {
     override fun goToMetadata() {
         val intent = Intent(this, MetadataActivity::class.java)
         startActivity(intent)
-        //Finish the activity to avoid keeping all the bitmaps in memory
-        finish()
```
jbors, line 111: Are you sure we should remove this? I am not sure how changing the bindings fixes the issue we had with keeping the Activities (and thus the Bitmaps) on the stack.

Also, when I take a heapdump I see multiple MaskActivities again

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataActivity.kt
```diff
@@ -34,37 +33,46 @@ class MetadataActivity : AppCompatActivity(), MetadataView, CaptureFlow {
...
-                LinearLayoutManager(this@MetadataActivity, LinearLayoutManager.HORIZONTAL, false)
-            adapter = imagesAdapter
-        }
-
-        lifecycleScope.launch {
-            presenter.showScreen()
-        }
+        binding = ActivityMetadataBinding.inflate(layoutInflater)
+        with(binding) {
+            setContentView(root)
```
jbors, line 44: This `root` feels a bit magic to me. But maybe that is because I am unfamiliar with android development.

Is there any way to refer to this that makes it more clear what exactly it is?

app/src/main/java/net/aiscope/gdd_app/ui/microscope_quality/MicroscopeQualityActivity.kt
```diff
@@ -29,38 +28,42 @@ class MicroscopeQualityActivity : AppCompatActivity(), MicroscopeQualityView, Ca
...
-        lifecycleScope.launch {
-            presenter.showScreen()
-        }
+        binding = ActivityMicroscopeQualityBinding.inflate(layoutInflater)
+        with(binding) {
+            setContentView(root)
 
-        setSupportActionBar(toolbar)
-        attachCaptureFlowToolbar(toolbar)
+            setSupportActionBar(toolbar.toolbar)
```
jbors, line 41: I am not quite sure what gets referenced by `toolbar.toolbar` either.

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataActivity.kt
eduardb: Not really, `root` is how the field is called in the parent `ViewBinding` class that comes with the AndroidX library. But I would say that `root` is quite a common way to refer to the top-most view in a layout, so I don't think that this should be confusing for a regular Android developer.

app/src/main/java/net/aiscope/gdd_app/ui/microscope_quality/MicroscopeQualityActivity.kt
eduardb: Fair enough. The first `toolbar` is the name of the field that points to the `ToolbarBinding` of the included `toolbar.xml` layout, whereas the second `toolbar` is the Toolbar itself in the included layout. The names come from the IDs that were used. 
I renamed the first `toolbar` to `toolbarLayout` to make it clearer, see 28e9b03a080649890b2e5d78efe9107c75a51432

app/src/main/java/net/aiscope/gdd_app/ui/mask/MaskActivity.kt
jbors: 👍

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataActivity.kt
jbors: OK fair enough

app/src/main/java/net/aiscope/gdd_app/ui/microscope_quality/MicroscopeQualityActivity.kt
jbors: 💯

app/build.gradle
jbors: OK far more work than I thought :/ Think it's an improvement overall though. Thanks!

PR #52, opened 2020-08-10
Add "comments" in sample metadata

mei4:


app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataPresenter.kt
```diff
@@ -53,12 +55,13 @@ class MetadataPresenter @Inject constructor(
         view.showInvalidFormError()
     }
 
-    suspend fun save(smearTypeId: Int, speciesValue: String) {
+    suspend fun save(smearTypeId: Int, speciesValue: String, comments: String?) {
```
eduardb, line 58: This method is never called with a `null` value for the `comments` parameter, why make it nullable?

app/src/main/res/layout/activity_metadata.xml
```diff
@@ -158,13 +158,49 @@
...
+                app:layout_constraintTop_toBottomOf="@+id/metadata_comments_title">
+
+                <EditText
+                    android:id="@+id/metadata_comments_input"
+                    android:layout_width="match_parent"
+                    android:layout_height="match_parent"
+                    android:ems="10"
+                    android:inputType="text"
+                    android:hint="@string/metadata_comments_input_hint" />
+            </com.google.android.material.textfield.TextInputLayout>
```
eduardb, line 195: Instead of using a `TextView` for the section title, and then a hint as `Input text` for the `TextInputLayout` (which personally, I think that it brings no value, as it doesn't tell you what the field is about), why not using the section title as the hint directly, and drop the `TextView`? This is just my opinion, but I would like to hear your thoughts. Maybe also run it with the designer too or someone else :D

mei4: The reason I wrapped the `EditText`is because I needed to add the helper text `Please, don't include any sensitive information.`.

I'm not sure if I understand your proposal. Do you mean moving the hint `Input text` and replacing it with the helper text `Please, don't include [...]`? Doing it like this we could remove the `TextInputLayout`. I agree that the `Input text` field does not bring any value.

eduardb: I mean the other way around, using the section title `@string/metadata_comments_title` as the hint. Something like this:
```suggestion
            <com.google.android.material.textfield.TextInputLayout
                android:id="@+id/metadata_comments_layout"
                android:layout_width="match_parent"
                android:layout_height="wrap_content"
                android:layout_marginStart="16dp"
                android:layout_marginEnd="16dp"
                android:textColorHint="@color/colorPlainText"
                app:helperText="@string/metadata_comments_title"
                app:helperTextEnabled="true"
                app:helperTextTextColor="@color/colorPlainText"
                app:layout_constraintEnd_toEndOf="parent"
                app:layout_constraintHorizontal_bias="0.0"
                app:layout_constraintTop_toBottomOf="@+id/metadata_species_spinner"
                app:layout_constraintVertical_bias="0.0">

                <EditText
                    android:id="@+id/metadata_comments_input"
                    android:layout_width="match_parent"
                    android:layout_height="match_parent"
                    android:ems="10"
                    android:inputType="text"
                    android:hint="@string/metadata_comments_input_hint" />
            </com.google.android.material.textfield.TextInputLayout>
```

mei4: Oh, I see, thanks for the clarification.

I applied it to see how it looks. I swapped the text, I think the helper looks good at the bottom

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataPresenter.kt
mei4: I'm not sure if in the `MetadataActivity` when we save and call `binding.metadataCommentsInput.text.toString()` if we could have a null pointer exception given that `.text` returns an `Editable!`. If that's the case I think it may make sense to leave the type as `String?`.

mei4:
> * In the `fillForm` method of the `MetadataActivity` class, shouldn't any previous comment be assigned to its corresponding `TextView`?
 
What I understood is that this method prepares the form with some preset data. In the case of the comments I don't think we need it. Please, let me know if I misunderstood the functionality of this method.

app/src/main/java/net/aiscope/gdd_app/ui/metadata/MetadataPresenter.kt
eduardb: Just found this: https://stackoverflow.com/questions/49561058/does-textview-gettext-ever-returns-null
If you want to be super defensive, you can default it to an empty string if for some weird reason the text is null, but I don't think that it's worth it.

eduardb:
> > * In the `fillForm` method of the `MetadataActivity` class, shouldn't any previous comment be assigned to its corresponding `TextView`?
> 
> What I understood is that this method prepares the form with some preset data. In the case of the comments I don't think we need it. Please, let me know if I misunderstood the functionality of this method.

Oh, you are right, I was conflating two different things. What I was actually thinking of is that if someone enters some comment and then add another image, when they are back on this screen the comment will be gone. Not a huge deal, but a bit inconvenient for sure. Maybe there is a way to prevent this kind of behaviour. If it's too complicated, we can leave it like this for now, I don't mind.

mei4:
> * In the `fillForm` method of the `MetadataActivity` class, shouldn't any previous comment be assigned to its corresponding `TextView`?

I pushed with these changes and also with what we discussed about the design.

eduardb:
It all looks good, but I realised that we are missing some tests for the mapping between the domain model and the DTOs. Would you be up for improving the coverage in this area?

Issue #21, opened 2020-01-17
Downscale images larger than the maximum bitmap size


eduardb:
In the mask layer, we are loading a large bitmap that comes from the camera. However, on some devices, this bitmap exceeds the maximum size that can be loaded in a canvas. For example, on a Redmi 6A device, the picture is 3120x4160 in size, whereas the maximum bitmap can be 4096x4096. 
As a solution, downscale the image so that it fits in the constraints.

See https://stackoverflow.com/questions/7428996/hw-accelerated-activity-how-to-get-opengl-texture-size-limit

jbors:
Yo @eduardb I _think_ this should be fixed now. Any QA you wanna do before closing the issue?

eduardb:
#55, #59 and #63 should close this issue, no need to do more QA apart from what we did already.

PR #55, opened 2020-08-25
[156] Downsample photos with too big a resolution

jbors:
Fix for https://github.com/theaiscope/GDD-app/issues/21

Downsample photo's until they are beneath the max canvas size


app/src/androidTest/java/net/aiscope/gdd_app/ui/metadata/SampleImagesAdapterTest.kt
```diff
@@ -0,0 +1,69 @@
...
+            decodeSampledBitmapAndCache(targetFile, 20, 20, cacheDir)
+        }
+
+        assertTrue(cacheDir.isDirectory)
+        val cachedFiles = cacheDir.listFiles()
+
+        assertTrue(cachedFiles.any { file -> file.name.equals("test_20x20.bmp", true) })
+        assertTrue(b.height < 50)
+        assertTrue(b.width < 50)
+        assertTrue(b.height > 20)
```
eduardb, line 43: Should we assert equality instead? Or do you think that this would couple the test too much to the implementation details? Either way, 20 would be a valid dimension as well. so this should be `>=`, no?

app/src/androidTest/java/net/aiscope/gdd_app/ui/metadata/SampleImagesAdapterTest.kt
```diff
@@ -0,0 +1,69 @@
...
+    private val targetContext = InstrumentationRegistry.getInstrumentation().targetContext
+    private val cacheDir = targetContext.cacheDir
+    private lateinit var targetFile: File
+
+    @Before
+    fun setup() {
+        targetFile = File(targetContext.filesDir, "test.bmp");
+        targetFile.createNewFile()
+
+        val stream = testContext.assets.open("w3c_home.bmp");
```
eduardb, line 26: In `MaskActivityTest.kt` there's a `getAssetStream` method that does this, and even has a failback. I think that it could be made an extension function on `Context`, and be reused here and in `BitmapReaderTest` as well, what do you think?

app/src/androidTest/java/net/aiscope/gdd_app/ui/metadata/SampleImagesAdapterTest.kt
```diff
@@ -0,0 +1,69 @@
...
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import java.io.File
+import java.io.InputStream
+
+@RunWith(AndroidJUnit4::class)
+class SampleImagesAdapterTest {
```
eduardb, line 15: The tests in this class don't really have to do anything with the adapter itself, but with some helper method that happens to be in the same file as the adapter. Thus, I am thinking, maybe it would make sense to move that method to its own file? Honestly, I wouldn't mind having it in the `BitmapReader` class, but whatever you prefer :)

app/src/main/java/net/aiscope/gdd_app/ui/metadata/SampleImagesAdapter.kt
```diff
@@ -86,73 +89,47 @@ private class ImageViewHolder(
         itemView.setOnClickListener { onImageClicked(imageFile, maskFile) }
     }
 
-    fun bind(image: File, mask: File) {
+    fun bind(image: File, mask: File, cacheDir: File) {
```
eduardb, line 92: `cacheDir` is not used, did you mean using it below instead of `itemView.context.cacheDir`? Otherwise, it can be removed.

app/src/main/java/net/aiscope/gdd_app/ui/metadata/SampleImagesAdapter.kt
```diff
@@ -86,73 +89,47 @@ private class ImageViewHolder(
...
-        // Calculate inSampleSize
-        inSampleSize = calculateInSampleSize(this, reqWidth, reqHeight)
-
-        // Decode bitmap with inSampleSize set
-        inJustDecodeBounds = false
-
-        val bitmap = BitmapFactory.decodeFile(image.absolutePath, this)
+    val bitmap = BitmapReader.decodeSampledBitmapFromResource(
+        image,
+        false,
```
eduardb, line 125: Would you use a named argument here, so it's clear what does the `false` stands for, please?

app/src/androidTest/java/net/aiscope/gdd_app/ui/metadata/SampleImagesAdapterTest.kt
jbors: I had it in BitmapReader initially but moved it back for a reason that now does not exist anymore 😆 

Probs best to move it back

app/src/main/java/net/aiscope/gdd_app/ui/metadata/SampleImagesAdapter.kt
jbors: Had to change the order since it's not allowed to mix positional & named arguments, but done

app/src/androidTest/java/net/aiscope/gdd_app/ui/metadata/SampleImagesAdapterTest.kt
jbors: Moved

app/src/androidTest/java/net/aiscope/gdd_app/ui/metadata/SampleImagesAdapterTest.kt
jbors: Well I don't want to pin myself down to a certain number as a result since that's not explicitly expected. You're right on the equality though, adjusted (even though it does not occur with this specific implementation ;))

app/src/androidTest/java/net/aiscope/gdd_app/ui/metadata/SampleImagesAdapterTest.kt
jbors: Done. Also moved the `ParcelableExt` to a new `test.extensions` package

app/src/androidTest/java/net/aiscope/gdd_app/ui/metadata/SampleImagesAdapterTest.kt
```diff
@@ -0,0 +1,69 @@
...
+
+        assertTrue(cachedFiles.any { file -> file.name.equals("test_20x20.bmp", true) })
+        assertTrue(b.height < 50)
+        assertTrue(b.width < 50)
+        assertTrue(b.height > 20)
+        assertTrue(b.width > 20)
+    }
+
+    @Test
+    fun shouldNotWriteToCacheTwice() {
```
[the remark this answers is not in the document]
jbors: So apparently the `lastmodified` is in seconds. So the test runs in <1 second and you get the same value on `lastmodified`

I added a Thread.sleep(2000) between the first and second calls, and then it will fail if you do not return the cached file. But that's a bit, ewww... Suggestions?

app/src/androidTest/java/net/aiscope/gdd_app/ui/metadata/SampleImagesAdapterTest.kt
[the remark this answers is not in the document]
eduardb: `Bitmap` is just how you hold in memory any kind of image, which doesn't matter on the file encoding. `BitmapFactory.decodeFile()` can decode both BMP and PNG, among others, without any extra effort required from your part :) Anyway, as you wish.

app/src/androidTest/java/net/aiscope/gdd_app/ui/metadata/SampleImagesAdapterTest.kt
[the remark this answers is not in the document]
eduardb: Would it work with less than 2 seconds? :D 
An alternative would be to maybe use a `WatchService` to see if the file has been modified, but not sure if it's worth it. Here is an article on how it would work: https://docs.oracle.com/javase/tutorial/essential/io/notification.html

app/src/androidTest/java/net/aiscope/gdd_app/ui/metadata/SampleImagesAdapterTest.kt
eduardb: Fair enough. Do you think that it would make sense to have a parametrised test here though? To catch more edge cases.

app/src/main/java/net/aiscope/gdd_app/ui/util/BitmapReader.kt
```diff
@@ -0,0 +1,94 @@
...
+        cacheDir: File
+    ): Bitmap = withContext(Dispatchers.IO) {
+        val cachedImage = File(
+            cacheDir,
+            "${image.nameWithoutExtension}_${reqWidth}x${reqHeight}.${image.extension}"
+        )
+        if (cachedImage.exists()) {
+            return@withContext BitmapFactory.decodeFile(cachedImage.absolutePath)
+        }
+        val bitmap = BitmapReader.decodeSampledBitmapFromResource(
```
eduardb, line 47: `BitmapReader` is redundant here :D

eduardb (approved):
Looks solid to me. I am approving the PR, feel free to apply or not my suggestions from my last comments, and merge ;)

app/src/androidTest/java/net/aiscope/gdd_app/ui/metadata/SampleImagesAdapterTest.kt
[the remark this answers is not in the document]
jbors: Hahaha well, bigger than 1 second at least. 1200 ms would be acceptable I guess :P 

I don't think the `WatchService` the best idea, I was thinking more along the lines of using Mockito to see if `writeToFile` gets called twice. Not _exactly_ the same test but similar enough I guess.

app/src/main/java/net/aiscope/gdd_app/ui/util/BitmapReader.kt
jbors: Ah yeah. Fixed

app/src/androidTest/java/net/aiscope/gdd_app/ui/metadata/SampleImagesAdapterTest.kt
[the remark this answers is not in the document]
jbors: Hmmmm yeah it gets read from a JPG in most cases anyways. OK adjusted

app/src/androidTest/java/net/aiscope/gdd_app/ui/metadata/SampleImagesAdapterTest.kt
jbors: Well these just test basic functionality. I am sure we _could_ add more tests for edge cases, but that sort of goes for the entire code base ;)

I don't think it's worth the effort right now

app/src/androidTest/java/net/aiscope/gdd_app/ui/metadata/SampleImagesAdapterTest.kt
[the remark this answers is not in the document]
jbors: Changed it to `1200ms` for now

PR #56, opened 2020-09-01
[154] Differentiaties masked images

mei4:
Regarding ticket 154 about differentiating masked images and non masked I have two concerns:
1- I’m not sure what’s the best way to check when a mask is empty. For now I only managed to do it checking the length of the file.
2- I’m struggling to replicate the design, I didn’t find a simple way to put the dot slightly outside the image without changing the image size.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -23,8 +23,8 @@ android {
         applicationId "net.aiscope.gdd_app"
         minSdkVersion 23
         targetSdkVersion 29
-        versionCode 24
-        versionName "2.1.4"
+        versionCode 25
+        versionName "2.1.5"
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
 
         buildConfigString 'SMARTLOOK_API_KEY', buildProperties.secrets['SMARTLOOK_API_KEY'].or(buildProperties.env['SMARTLOOK_API_KEY']).or("")


app/src/main/java/net/aiscope/gdd_app/ui/metadata/SampleImagesAdapter.kt
```diff
@@ -46,7 +59,11 @@ class SampleImagesAdapter(
     override fun getItemViewType(position: Int): Int {
         return when (position) {
             0 -> R.layout.item_metadata_add_image
-            else -> R.layout.item_metadata_sample_image
+            else -> {
+                if (masks[position - 1].length().toInt() > EMPTY_MASK_LENGTH)
```
eduardb, line 63: You shouldn't need two different layouts/view types for this. You can use only one layout, where you have the dot, and then just hide or show the dot view in the `bind()` method of the view holder, based on the business logic.

app/src/main/java/net/aiscope/gdd_app/ui/metadata/SampleImagesAdapter.kt
```diff
@@ -23,14 +23,27 @@ class SampleImagesAdapter(
 ) :
     RecyclerView.Adapter<RecyclerView.ViewHolder>() {
 
+    companion object {
+        private const val EMPTY_MASK_LENGTH = 4885
```
eduardb, line 27: I imagine this size is dependent on the device, so this specific value won't work on a wide range of devices.
Like discussed on Slack, let's try to detect in the `MaskActivity` whether the user has drawn anything or not, and save that as a flag in the `Sample` object.

app/src/main/java/net/aiscope/gdd_app/ui/metadata/SampleImagesAdapter.kt
```diff
@@ -90,15 +107,17 @@ private class ImageViewHolder(
         imageFile = image
         maskFile = mask
 
+        val img: ImageView = itemView.findViewById(R.id.sample_image)
```
eduardb, line 110: Let's try to use [View Binding](https://developer.android.com/topic/libraries/view-binding) instead of `findViewById`. Also, this should be done only once when the view holder is instantiated, not on each `bind`, as the layout is not going to chance ;)

app/src/main/res/layout/item_metadata_add_image.xml
```diff
@@ -1,26 +1,40 @@
...
-    android:orientation="vertical">
 
-    <ImageView
-        android:layout_width="wrap_content"
-        android:layout_height="wrap_content"
-        android:paddingStart="1dp"
-        android:src="@drawable/ic_add_photo_rounded"
-        tools:ignore="RtlSymmetry"
-        android:importantForAccessibility="no" />
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
eduardb, line 3: I am not sure why did you change this layout. Also, there's no point in using a `ConstraintLayout` if you only have one child.

app/src/main/res/layout/item_metadata_sample_image_with_mask.xml
```diff
@@ -0,0 +1,47 @@
...
+    android:layout_marginStart="3dp"
+    android:layout_height="wrap_content"
+    android:layout_width="wrap_content"
+    android:clipChildren="false"
+    android:clipToPadding="false"
+    android:paddingTop="6dp"
+    android:paddingEnd="5dp"
+    android:paddingStart="0dp">
+
+    <androidx.constraintlayout.widget.ConstraintLayout
```
eduardb, line 15: Also, there's no need for needing a `ConstraintLayout` inside another `ConstraintLayout`. You can achieve whatever needed with only one.

app/src/main/res/layout/item_metadata_sample_image_with_mask.xml
```diff
@@ -0,0 +1,47 @@
...
+            tools:src="@drawable/ic_add_photo_rounded"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintBottom_toBottomOf="parent"/>
+
+        <ImageView
+            android:id="@+id/mask_dot"
+            android:layout_width="16dp"
+            android:layout_height="16dp"
+            android:src="@drawable/ic_dot"
+            android:contentDescription="@string/metadata_sample_mask_dot_description"
```
eduardb, line 38: 💯 for accessibility :)

eduardb (approved):
Besides rethinking the masked images detection, please optimize the use of layout resources and `ConstraintLayout`s

app/src/main/java/net/aiscope/gdd_app/ui/metadata/SampleImagesAdapter.kt
mei4: I haven't thought about this and it makes a lot of sense. I'll implement it

app/src/main/java/net/aiscope/gdd_app/ui/metadata/SampleImagesAdapter.kt
mei4: I'm working on this issue. I still haven't push it but what I do now is check that the captured mask is an empty bitmap

app/src/main/res/layout/item_metadata_add_image.xml
mei4: I did it to align the items inside the RecyclerView in the metadata activity. When I added the dot the add sample item was sticked to the top and I could find how to move it to the bottom

app/src/main/res/layout/item_metadata_sample_image.xml
```diff
@@ -1,10 +1,39 @@
...
+        android:contentDescription="@string/metadata_sample_mask_dot_description"
+        android:visibility="invisible"
+        app:layout_constraintTop_toTopOf="parent"
+        app:layout_constraintBottom_toTopOf="parent"
+        app:layout_constraintStart_toEndOf="parent"
+        app:layout_constraintEnd_toEndOf="parent" />
+
+</androidx.constraintlayout.widget.ConstraintLayout>
+
+
```
eduardb, line 39: More straightforward to be done something like this:
```xml
<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
    xmlns:tools="http://schemas.android.com/tools"
    android:layout_width="wrap_content"
    android:layout_gravity="top"
    android:layout_marginEnd="2dp"
    tools:ignore="RtlSymmetry">

    <ImageView
        android:id="@+id/sample_image"
        android:layout_width="@dimen/sample_image_thumbnail_width"
        android:layout_height="@dimen/sample_image_thumbnail_height"
        android:layout_gravity="bottom|start"
        android:layout_marginTop="@dimen/sample_image_dot_displacement"
        android:layout_marginEnd="@dimen/sample_image_dot_displacement"
        android:background="@drawable/item_metadata_add_image_background"
        android:contentDescription="@string/metadata_sample_image_content_description"
        android:scaleType="centerCrop"
        tools:src="@drawable/ic_add_photo_rounded" />

    <ImageView
        android:id="@+id/mask_dot"
        android:layout_width="16dp"
        android:layout_height="16dp"
        android:layout_gravity="top|end"
        android:contentDescription="@string/metadata_sample_mask_dot_description"
        android:src="@drawable/ic_dot"
        android:visibility="gone"
        tools:visibility="visible" />

</FrameLayout>
```

eduardb (changes requested):
The adapter needs some things fixed. I also added some suggestions about domain modelling. 
To speed things up on the layout side, I added comments with a working implementation. You just need to add `<dimen name="sample_image_dot_displacement">6dp</dimen>` in `dimens.xml`.

app/src/main/java/net/aiscope/gdd_app/model/SampleModels.kt
```diff
@@ -14,18 +14,26 @@ data class Sample(
     val microscopeQuality: MicroscopeQuality? = null,
     val images: LinkedHashSet<File> = linkedSetOf(),
     val masks: LinkedHashSet<File> = linkedSetOf(),
+    val hasMask: MutableList<Boolean> = mutableListOf(),
```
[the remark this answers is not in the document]
mei4: I think it makes a lot a sense. I'll work on it!

eduardb:
I messed up with a git command and the PR was closed automatically. I will open a new one instead, sorry!

PR #65, opened 2020-09-27
[154] Differentiate masked images

eduardb:
This is just PR #56 opened again and continued by me

| Before | After |
|--------|-------|
| device-2020-10-01-152121 | device-2020-10-01-151609 |

**Note for reviewers**
I introduced some _improvements_ of the domain modelling of the images in the `Sample` class. Instead of having different lists for the blood images, masks and now for the flags of whether the masks are empty or not, now we have an `Image` immutable class that contains an optional `InProgressCapture` object that is initialised with the blood image before it gets masked, and an array of `CompletedCapture` objects. This approach should also fix the errors we saw with unequal number of images and masks.


app/src/main/java/net/aiscope/gdd_app/repository/SampleDtos.kt
```diff
@@ -22,25 +24,42 @@ data class SampleDto(
...
+            images = Images(inProgressCapture, completedCaptures),
+            metadata = metadata.toDomain(),
+            status = SampleStatus.values().first { it.id == status },
+            createdOn = createdOn,
+            lastModified = lastModified
+        )
+    }
+
+    private fun backfillAreMaskEmpty() =
+        if (areMasksEmpty.isNotEmpty()) areMasksEmpty else List(imagePaths.size) { false }
```
jbors, line 54: So this backfill is only for samples which have been made on a previous version of the app? Or is this load bearing?

app/src/main/java/net/aiscope/gdd_app/ui/mask/MaskActivity.kt
```diff
@@ -190,4 +192,11 @@ class MaskActivity : AppCompatActivity(), MaskView, CaptureFlow {
             mutable = mutable
         )
     }
+
+    private fun isEmptyMaskBitmap(): Boolean {
+        val maskBitmap = binding.photoMaskView.getMaskBitmap()
+        val emptyBitmap: Bitmap =
+            Bitmap.createBitmap(maskBitmap.width, maskBitmap.height, maskBitmap.config);
```
jbors, line 199: Seems a bit wasteful to create this on every call to the method. You reckon the width/height/config may change between samples?

app/src/main/java/net/aiscope/gdd_app/model/SampleModels.kt
```diff
@@ -56,6 +53,54 @@ data class MicroscopeQuality(
     val magnification: Int
 )
 
+data class Images(
```
jbors, line 56: So this `Images` class consists of a list of captures. Wouldn't it be better to name it something like `Captures` or `CaptureList`? It confused me a bit when reading code in other classes

app/src/main/java/net/aiscope/gdd_app/model/SampleModels.kt
```diff
@@ -56,6 +53,54 @@ data class MicroscopeQuality(
...
+                )
+            inProgressCapture != null ->
+                Images(
+                    null,
+                    completedCaptures + CompletedCapture(
+                        inProgressCapture.image, maskPath, isEmpty
+                    )
+                )
+            else ->
+                Images(
```
jbors, line 83: Sorry, what is this use case exactly?

jbors (approved):
Looks good overall! 👍 

Added a few comments/questions but approved in advance

app/src/main/java/net/aiscope/gdd_app/repository/SampleDtos.kt
eduardb: Yes, it's for the former. The edge-case that this tries to avoid is when the app is updated to this new version when there are still samples to upload. Without this backfill, due to the `zip` below, we would lose the images for upload when recovering the data from the repository.

app/src/main/java/net/aiscope/gdd_app/ui/mask/MaskActivity.kt
eduardb: Good point. On the other hand, you probably don't want this large bitmap linger in memory after we are done with the comparison. Ideally, we would have some in-memory LRU cache for bitmaps, which we could also use for other parts where we use (disk) caching. Maybe this would be a good issue to open for #hacktoberfest :D

app/src/main/java/net/aiscope/gdd_app/ui/util/BitmapReader.kt
```diff
@@ -146,3 +145,11 @@ suspend fun calculateInSampleSize(
     }
     return@withContext inSampleSize
 }
+
+private fun File.bitmapCompressFormatFromExtension(): Bitmap.CompressFormat {
+    return when (this.extension.toLowerCase(Locale.US)) {
+        "png" -> Bitmap.CompressFormat.PNG
+        "webp" -> Bitmap.CompressFormat.WEBP
```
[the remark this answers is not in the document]
eduardb: No, we don't. I just wanted to be exhaustive here. Especially if we ever decide using WebP, I am 99% sure that we would miss updating this method, haha.

app/src/main/java/net/aiscope/gdd_app/model/SampleModels.kt
eduardb: I agree, renamed in b66dff67

app/src/main/java/net/aiscope/gdd_app/model/SampleModels.kt
eduardb: This branch is executed when editing the mask of a previous "capture". The code is not very clear about that, so I added some tests which I missed adding in the first place, to serve as documentation, and also did some refactoring and added some comments. Let me know what you think now, probably there is room for more improvement. See 91481252.

eduardb (commented):
Addressing comments :D

app/src/main/java/net/aiscope/gdd_app/ui/util/BitmapReader.kt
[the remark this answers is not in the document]
jbors: Fair enough 👍

app/src/main/java/net/aiscope/gdd_app/repository/SampleDtos.kt
jbors: Cool 👍

app/src/main/java/net/aiscope/gdd_app/ui/mask/MaskActivity.kt
jbors: Aight let's call this out-of-scope :)

jbors (approved):
LGTM

PR #68, opened 2020-10-07
[159] Do not process draw events until initialized

jbors:


app/src/androidTest/java/net/aiscope/gdd_app/ui/mask/customview/MaskCustomViewTest.kt
```diff
@@ -0,0 +1,74 @@
...
+import androidx.test.platform.app.InstrumentationRegistry
+import net.aiscope.gdd_app.test.extensions.getAssetStream
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Test
+import java.io.File
+import java.io.InputStream
+
+class MaskCustomViewTest{
```
eduardb, line 16: 👏

app/src/main/java/net/aiscope/gdd_app/ui/mask/customview/MaskCustomView.kt
```diff
@@ -53,11 +53,13 @@ class MaskCustomView @JvmOverloads constructor(
 
     @SuppressLint("ClickableViewAccessibility")
     override fun onTouchEvent(event: MotionEvent): Boolean =
-        super.onTouchEvent(event) ||
-                when (currentMode) {
-                    Mode.Zoom -> onTouchMove(event)
-                    Mode.Draw, Mode.Erase -> onTouchDraw(event)
-                }
+        this::drawableSize.isInitialized && (
```
eduardb, line 56: This definitely does the trick, but I feel a bit like it's treating the symptom, not the cause. How would you feel about replacing the condition with checking that the drawable has been set?
```suggestion
        this.drawable != null && (
```

eduardb (approved):
Looks great, just one small suggestion from my side.

app/src/main/java/net/aiscope/gdd_app/ui/mask/customview/MaskCustomView.kt
jbors: Hmmm. Actually I think I disagree. `drawableSize` is contained in this class and actually used in the `onTouchDraw` method. 

`drawable` is inherited from `ImageView` and not directly used within this class. 

To me it seems preferable to use `drawableSize`

jbors: @eduardb should we introduce a `readyForEvents` boolean that allows us to switch responding to events on and off from code?

eduardb: Hmm, how would that look like? If it's not increasing the complexity too much, sure. Otherwise, let's just stick with the `drawableSize` here.

PR #80, opened 2020-10-08
Follow-up updates

eduardb:
Updates the following:
 - remaining libraries and plugins
 - gradle
 - bitrise steps

Removes some code that became deprecated with the update to Kotlin 1.4 and uses SAM conversion where possible


jbors (approved):
👍

PR #90, opened 2020-10-23
Update firebase functions

jbors:
We had a vulnerability so I updated the firebase functions to the latest version

image


CeasarCyrillus (approved):

eduardb:
@jbors have you tried deploying the Firebase function with the updated dependencies?

PR #105, opened 2020-11-16
Make sure to only take 1 picture

jbors:
Since a microscopist can also take a picture with the 'volume up' button, it is quite possible to accidentally take a bunch. This potentially causes crashes. 

Changed to only process 1 image capture & then go to the masking activity

### DIFF
--- a/app/src/main/java/net/aiscope/gdd_app/ui/capture/CaptureImagePresenter.kt
+++ b/app/src/main/java/net/aiscope/gdd_app/ui/capture/CaptureImagePresenter.kt
@@ -6,7 +6,7 @@ import timber.log.Timber
 
 class CaptureImagePresenter(
     val view: CaptureImageView,
-    val repository: SampleRepository,
+    val repository: SampleRepository
 ) {
     var processingImageCapture: Boolean = false
 
@@ -37,5 +37,7 @@ class CaptureImagePresenter(
 
     fun onCaptureError(it: CameraException) {
         Timber.tag("Camera Error").e(it, "capture error")
+        //Allow new photo to be taken on camera error
+        processingImageCapture = false
     }
 }
--- a/app/src/main/java/net/aiscope/gdd_app/ui/capture/CaptureImagePresenter.kt
+++ b/app/src/main/java/net/aiscope/gdd_app/ui/capture/CaptureImagePresenter.kt
@@ -6,20 +6,26 @@ import timber.log.Timber
 
 class CaptureImagePresenter(
     val view: CaptureImageView,
-    val repository: SampleRepository
+    val repository: SampleRepository,
+    var receivedImageCapture: Boolean = false
 ) {
 
     fun handleCaptureImageButton(imageName: String) {
-        Timber.tag("Taking Photo").d("button pressed")
-        view.takePhoto(imageName) { file ->
-            Timber.tag("Taking Photo").d(file?.absolutePath)
-            if (file == null) {
-                view.notifyImageCouldNotBeTaken()
-            } else {
-                val sample = repository.current().addNewlyCapturedImage(file)
-                repository.store(sample)
+        if (receivedImageCapture){
+            Timber.tag("Taking Photo").d("Already processing photo")
+        } else {
+            receivedImageCapture = true
+            Timber.tag("Taking Photo").d("button pressed")
+            view.takePhoto(imageName) { file ->
+                Timber.tag("Taking Photo").d(file?.absolutePath)
+                if (file == null) {
+                    view.notifyImageCouldNotBeTaken()
+                } else {
+                    val sample = repository.current().addNewlyCapturedImage(file)
+                    repository.store(sample)
 
-                view.goToMask(sample.disease, file.absolutePath, sample.nextMaskName())
+                    view.goToMask(sample.disease, file.absolutePath, sample.nextMaskName())
+                }
             }
         }
     }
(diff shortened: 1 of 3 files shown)


app/src/main/java/net/aiscope/gdd_app/ui/capture/CaptureImagePresenter.kt
```diff
@@ -6,20 +6,26 @@ import timber.log.Timber
...
-        view.takePhoto(imageName) { file ->
-            Timber.tag("Taking Photo").d(file?.absolutePath)
-            if (file == null) {
-                view.notifyImageCouldNotBeTaken()
-            } else {
-                val sample = repository.current().addNewlyCapturedImage(file)
-                repository.store(sample)
+        if (receivedImageCapture){
+            Timber.tag("Taking Photo").d("Already processing photo")
+        } else {
```
eduardb, line 16: You can do a `return` in the `if`, so you don't need the `else` (results in a smaller diff 😛 )

app/src/main/java/net/aiscope/gdd_app/ui/capture/CaptureImagePresenter.kt
```diff
@@ -6,20 +6,26 @@ import timber.log.Timber
...
+            receivedImageCapture = true
+            Timber.tag("Taking Photo").d("button pressed")
+            view.takePhoto(imageName) { file ->
+                Timber.tag("Taking Photo").d(file?.absolutePath)
```
eduardb, line 20: Shouldn't `receivedImageCapture` be set to false somewhere here? Especially if there was an error, we want to let the user to be able to try again.

eduardb (commented):
A couple of small comments, the second one is more important IMO.

eduardb:
Also, can this be covered by a test somehow?

app/src/main/java/net/aiscope/gdd_app/ui/capture/CaptureImagePresenter.kt
jbors: Personally not a big fan of using `return` as a break statement but sure

app/src/main/java/net/aiscope/gdd_app/ui/capture/CaptureImagePresenter.kt
jbors: Good point. Not quite sure to do it in or outside the `suspend onPhotoReceived` though.  I mean, ideally we wait for that to complete. It's also possible we never make it to that, either cause the `takePhoto` does not succeed here: https://github.com/theaiscope/GDD-app/blob/cbfecac1e7ed1e896c2e2e0f48dd725b1ff9a603/app/src/main/java/net/aiscope/gdd_app/ui/capture/CaptureImageActivity.kt#L97

or if something goes wrong in the `onPhotoReceived` itself

app/src/main/java/net/aiscope/gdd_app/ui/capture/CaptureImagePresenter.kt
```diff
@@ -6,10 +6,16 @@ import timber.log.Timber
 
 class CaptureImagePresenter(
     val view: CaptureImageView,
-    val repository: SampleRepository
+    val repository: SampleRepository,
+    var processingImageCapture: Boolean = false
```
eduardb, line 10: I don't really see a reason why this should be here in the constructor, how about making it a field?

app/src/main/java/net/aiscope/gdd_app/ui/capture/CaptureImagePresenter.kt
```diff
@@ -21,6 +27,8 @@ class CaptureImagePresenter(
 
                 view.goToMask(sample.disease, file.absolutePath, sample.nextMaskName())
             }
+            //What if shit falls over tho??
```
eduardb, line 30: Language, mister! xD

app/src/main/java/net/aiscope/gdd_app/ui/capture/CaptureImagePresenter.kt
```diff
@@ -6,10 +6,16 @@ import timber.log.Timber
...
+    var processingImageCapture: Boolean = false
 ) {
 
     fun handleCaptureImageButton(imageName: String) {
+        if (processingImageCapture){
+            Timber.tag("Taking Photo").d("Already processing photo")
+            return
+        }
+        processingImageCapture = true
         Timber.tag("Taking Photo").d("button pressed")
```
eduardb, line 19: Hmm, should we move this log above the `if`, for debugging crashes?

jbors: Sure

app/src/main/java/net/aiscope/gdd_app/ui/capture/CaptureImagePresenter.kt
jbors: Hahaha temp remark bro. Promise it won't make it into a release 🤞

app/src/main/java/net/aiscope/gdd_app/ui/capture/CaptureImagePresenter.kt
jbors: Sure

app/src/main/java/net/aiscope/gdd_app/ui/capture/CaptureImagePresenter.kt
eduardb: I know, I know, I was just messing with you :P

eduardb (approved):
looking good!

PR #134, opened 2021-01-10
[143] Design improvements

jbors:
- Palette changed for virus icon
- The color of the virus indicates the draw color
- Removed the dot behind the pen
- Header name updates


app/src/main/java/net/aiscope/gdd_app/ui/mask/MaskActivity.kt
```diff
@@ -175,10 +176,14 @@ class MaskActivity : AppCompatActivity(), MaskView, CaptureFlow {
     }
 
     private fun refreshBrushDrawableColor() {
+        // Retrieve the dot behind the draw icon and set to currently selected colour
         val stageBtnDrawable = binding.drawBtn.compoundDrawables[0] as LayerDrawable
```
eduardb, line 180: Do we still need to tint the draw button now that we tint the stages one? Seems redundant, and like a visual cacophony.

app/src/main/res/values/strings.xml
```diff
@@ -42,10 +42,10 @@
     <string name="microscope_quality_snackbar_error">An error occurred while saving your microscope quality data. Please try again.</string>
     <string name="microscope_quality_snackbar_retry">TRY AGAIN</string>
     <!-- CaptureImage -->
-    <string name="capture_activity_title">Capture photo</string>
+    <string name="capture_activity_title">Capture a sample</string>
```
eduardb, line 45: I am not sure this is correct though. A sample contains multiple captured photos/images, according to our domain entities: https://github.com/theaiscope/GDD-app/blob/20117b43e48c68bc6c9c21420fe450161c4419e6/app/src/main/java/net/aiscope/gdd_app/model/SampleModels.kt

eduardb (commented):
I dig the new virus icon :D But I have two questions, maybe we can make this even better.

jbors:
> I dig the new virus icon :D But I have two questions, maybe we can make this even better.

Dunno man, this is how it came from Irina & Lorenzo. There's a bunch more changes coming. Will ask them to have a look but maybe we should set up a short call to go through the new design?

eduardb:
> > I dig the new virus icon :D But I have two questions, maybe we can make this even better.
> 
> @eduardb Dunno man, this is how it came from Irina & Lorenzo. There's a bunch more changes coming. Will ask them to have a look but maybe we should set up a short call to go through the new design?

I see. I would prefer some async form of communication though, if possible, like Slack.

IriOva:
> > > I dig the new virus icon :D But I have two questions, maybe we can make this even better.
> > 
> > 
> > @eduardb Dunno man, this is how it came from Irina & Lorenzo. There's a bunch more changes coming. Will ask them to have a look but maybe we should set up a short call to go through the new design?
> 
> I see. I would prefer some async form of communication though, if possible, like Slack.

First time I comment something here :). Can we talk via Slack?

eduardb:
@IriOva sure, I have Slack open pretty much all the time, just ping me whenever works best for you.

app/src/main/res/values/strings.xml
jbors: 'image' now

app/src/main/java/net/aiscope/gdd_app/ui/mask/MaskActivity.kt
jbors: Dot is gone now

PR #142, opened 2021-01-22
[143] Move sample data into tabs

jbors:
Not all parts of the new design are in this yet, this just covers the tabbed functionality


app/src/main/java/net/aiscope/gdd_app/ui/sample_completion/SampleCompletionActivity.kt
```diff
@@ -0,0 +1,130 @@
...
+            val preparationOK = preparationFragment?.validateAndUpdateVM() ?: true
+
+            val qualityFragment: QualityFragment? =
+                findFragment(2)
+                        as? QualityFragment
+
+            val qualityOK = qualityFragment?.validateAndUpdateVM() ?: true
+
+            //Saving should only happen if all tabs are OK
+            return (preparationOK && qualityOK);
```
simtop, line 85: You don't need the ;

app/src/main/java/net/aiscope/gdd_app/ui/sample_completion/SampleCompletionViewModel.kt
```diff
@@ -0,0 +1,105 @@
...
+                status = SampleStatus.ReadyToUpload
+            )
+            val storedSample = repository.store(updatedSample)
+
+            //Put it in line for uploading to firebase
+            remoteStorage.enqueue(storedSample, context)
+        }
+    }
+
+    private fun getWaterType(waterTypeValue: String): WaterType {
```
simtop, line 83: It would be easier to manage if you have the strings inside the water type and use a enum with a string.

enum class WaterType(val name: Int) {
   DISTILLED(R.string.water_type_distilled), 
    GOLD("yellow"),
BOTTLED(R.string.water_type_bottled), 
   ...
}

app/src/main/java/net/aiscope/gdd_app/ui/sample_completion/SampleCompletionViewModel.kt
```diff
@@ -0,0 +1,105 @@
...
+        return when (waterTypeValue) {
+            context.getString(R.string.water_type_distilled) -> WaterType.DISTILLED
+            context.getString(R.string.water_type_bottled) -> WaterType.BOTTLED
+            context.getString(R.string.water_type_tap) -> WaterType.TAP
+            context.getString(R.string.water_type_well) -> WaterType.WELL
+            else -> throw IllegalStateException("$waterTypeValue water type is unknown")
```
simtop, line 89: Also whoever uses this needs a try catch

simtop (approved):

app/src/main/java/net/aiscope/gdd_app/ui/sample_completion/SampleCompletionActivity.kt
jbors: 👍

app/src/main/java/net/aiscope/gdd_app/ui/sample_completion/SampleCompletionViewModel.kt
jbors: True. I'm only moving existing functionality/code into the tabs though so I prefer not to change functions if I can avoid it. Will make a note for later

app/src/main/java/net/aiscope/gdd_app/ui/sample_completion/SampleCompletionViewModel.kt
```diff
@@ -0,0 +1,179 @@
...
+        viewModelScope.launch(Dispatchers.IO) {
+
+            val sample = repository.current()
+            disease = sample.disease
+            captures = sample.captures.completedCaptures
+
+            //So what we gonna do here for the meta stuff?
+            val lastMeta = repository.lastSaved()?.metadata
+            Timber.i("LAST META %s", lastMeta)
+            smearTypeId = MetadataMapper.getSmearTypeId(lastMeta?.smearType!!)
```
simtop, line 73: Hello, it's better not to use !! as much as posible because if it's null the app will crash. You have two options, or control the value with a if( != null) or a let?. or another option is to Modify the mappers to accept a null input, and control in the mapper that in the case of null set a default value.

app/src/main/java/net/aiscope/gdd_app/ui/sample_completion/SampleCompletionViewModel.kt
```diff
@@ -0,0 +1,179 @@
...
+            speciesValue = MetadataMapper.getSpeciesValue(context, lastMeta?.species!!)
```
simtop, line 74: same here

app/src/main/java/net/aiscope/gdd_app/ui/sample_completion/SampleCompletionViewModel.kt
```diff
@@ -0,0 +1,179 @@
...
+                giemsaFP,
+                usesPbs,
+                reusesSlides,
+                getBloodQuality(bloodQuality)
+            )
+
+            //Aight so what else for the meta stuff?
+            val newMeta = SampleMetadata(
+                //TODO: move these methods out of MDMapper
+                smearType = MetadataMapper.getSmearType(smearTypeId!!),
```
simtop, line 111: same here and the line under this.

app/src/main/res/layout/fragment_metadata.xml
```diff
@@ -1,16 +1,183 @@
 <?xml version="1.0" encoding="utf-8"?>
-<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<LinearLayout
```
simtop, line 2: Why is this inside a linear layout?

jbors: Cause it wat a linear layout before ;)

Like said before, I prefer moving over changing/improving right now. This task is big enough as it is

app/src/main/java/net/aiscope/gdd_app/ui/sample_completion/SampleCompletionViewModel.kt
jbors: Changed the mapper to take nullable types

app/src/main/java/net/aiscope/gdd_app/ui/sample_completion/SampleCompletionViewModel.kt
jbors: Changed the mapper to take nullable types

app/src/main/java/net/aiscope/gdd_app/ui/sample_completion/SampleCompletionViewModel.kt
jbors: Changed the mapper to take nullable types

app/src/main/java/net/aiscope/gdd_app/dagger/SampleCompletionModule.kt
```diff
@@ -0,0 +1,55 @@
...
+            SampleCompletionViewModel(repository, remoteStorage, context)
+    }
+}
+
+@Singleton
+class SampleCompletionViewModelFactory @Inject constructor(
+    private val provider: Provider<SampleCompletionViewModel>
+) : ViewModelProvider.Factory {
+
+    override fun <T : ViewModel> create(modelClass: Class<T>): T {
```
rkunal, line 46: Can create() be overridden in this way ?
```
override fun <T : ViewModel> create(SampleCompletionViewModel: Class<T>): T {
        return provider.get() as T
    }
```

[Reference Doc](https://developer.android.com/reference/androidx/lifecycle/ViewModelProvider.Factory#create(java.lang.Class%3CT%3E))

jbors: I mean, maybe this is some Kotlin thingy I am unfamiliar with, but that would not prevent you from calling the method with a different class right?

`SampleCompletionModel` simply becomes the new name for 'modelClass' I think (And a pretty confusing one at that)

rkunal: You are right. Embarrassingly, I got confused with an unfamiliar syntax.

PR #169, opened 2021-03-19
[166] - removed "Uses alcohol" toggle

egovova11:
- removed "Uses alcohol to fix the sample" toggle (166)


eduardb (approved):
Removing code is the best kind of PR 🎉

PR #170, opened 2021-03-19
[163] - ensure we are using the highest resolution when doing shots

egovova11:
- explicitly use the highest resolution when taking a photo


eduardb (approved):
Looks great!

PR #195, opened 2021-05-09
143 tabs

simtop:
The viewbinding can leak in fragments, so you have to assign a null when the fragment is destroyed.

A recommended way to use the coroutines inside the viewmodel is using viewmodel scope, the advantage it gives us is that it's lifecycle aware so the viewmodel will be the responsible for the suspended function it calls.

I wrote a TODO, because its a good practice to inject the coroutines dispatcher so you can change it for the test one while testing, but I will comment it in a meeting to see if we want to do that.


jbors:
Added to https://github.com/theaiscope/GDD-app/pull/142 manually

PR #239, opened 2021-07-24
173 first time sample form - Change form/pager behaviour for different use cases

rkunal:
Changes :-
- Defined 2 Behaviours for Sample Form
- Store the flag for "first time storing form" in Microscopist model , along with "hasAcceptedPrivacyPolicy"
- new Interface for fragments inside Sample Form Activity

Behaviours Added:-
1.  When the user submits the form first time
    1. Submit-button label is "Continue"
    2. On-click of submit-button - move user to next tab. 
    3.  Submit-button label is "Save Sample" on last tab
    4. On-click of submit-button  on last tab -  save data
   
2. When the user has already submitted the form once in the past
    1. Keep old behaviour. ie. Show Submit button label "Save" and allow saving on any tab.

😅 🤞


app/src/main/java/net/aiscope/gdd_app/ui/sample_completion/SampleCompletionActivity.kt
```diff
@@ -105,52 +120,75 @@ class SampleCompletionActivity : CaptureFlow, AppCompatActivity() {
...
+    fun saveToVM() {
             try {
                 sharedVM.save()
-                finishFlow()
+                goToHomeAndConfirmSaved()
             } catch (@Suppress("TooGenericExceptionCaught") error: Throwable) {
                 Timber.e(error, "An error occurred when saving sample completion data")
-                showRetryBar()
+                //showRetryBar()
+                // Fn Moved here to temporary fix TooManyFunctions ¯\_(ツ)_/¯
```
jbors, line 139: I mean, if we are just moving stuff cause of the 'toomanyfunctions' I'd rather we just suppress the rule

rkunal: I will move them back and add the suppress

simtop (approved):

PR #241, opened 2021-08-02
175 Delete Sample Image

rkunal:
1. Menu button to delete sample image 
2. launch activity based on number of images left in sample

### DIFF
--- a/app/src/main/java/net/aiscope/gdd_app/ui/mask/MaskActivity.kt
+++ b/app/src/main/java/net/aiscope/gdd_app/ui/mask/MaskActivity.kt
@@ -127,17 +127,17 @@ class MaskActivity : AppCompatActivity(), MaskView, CaptureFlow {
         }
     }
 
-    private fun <T> T.showConfirmImageDeleteDialog(sample: Sample, file: File) where T : AppCompatActivity, T : CaptureFlow {
+    private fun <T> T.showConfirmImageDeleteDialog(
+        sample: Sample, file: File
+    ) where T : AppCompatActivity, T : CaptureFlow {
         with(AlertDialog.Builder(this, R.style.Theme_AiScope_Dialog)) {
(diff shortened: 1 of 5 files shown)


app/src/main/java/net/aiscope/gdd_app/ui/mask/MaskActivity.kt
```diff
@@ -104,6 +109,45 @@ class MaskActivity : AppCompatActivity(), MaskView, CaptureFlow {
...
+    override fun onCreateOptionsMenu(menu: Menu?): Boolean {
+        menuInflater.inflate(R.menu.mask_menu, menu)
+        return true
+    }
+
+    override fun onOptionsItemSelected(item: MenuItem): Boolean {
+        return when(item.itemId) {
+            R.id.action_delete_image -> {
+                lifecycleScope.launch {
+                    val file = File(checkNotNull(intent.getStringExtra(EXTRA_IMAGE_NAME)))
```
simtop, line 121: This could throw an exception, would be nice to check it outside of the File.

simtop (approved):

app/src/main/java/net/aiscope/gdd_app/ui/mask/MaskActivity.kt
rkunal: True. I will add the catch and show appropriate message to user.

I have 1 doubt. @simtop  
_File_ constructor only throws _NullPointerException_. And _checkNotNull_ throws  _IllegalStateException_ . I dont think _NullPointerException_ will ever be thrown when _checkNotNull_ exists.

Should I :-
1. Remove _checkNotNull_ and check for _File_'s Exception ? OR
2. Check for _checkNotNull_'s exception ?

rkunal: I have made the changes. Please feel free to open the conversation for any improvements.

PR #294, opened 2021-10-31
Demo branch from  old build. Store sample data , whenever SampleCompletionActivity is closed

rkunal:
_This is not a PR for merge_

Changes are previewed in this PR. Some TODOs added in codebase to share concerns.

whenever SampleCompletionActivity is closed  :-
1. Store data in SharedPreferences, without sending to remote
2. Allow storing data without validation, when not sending to remote
3. Add new enum as default values for various properties, when storing without validation

Pros : 
1. Maintains current architecture without introducing new data storage 
2. Avoids passing around of intent during user navigation

Cons :
1. Data validation responsibility belongs to UI. Can be added to model layer too, but will need work

### DIFF
--- a/app/src/main/java/net/aiscope/gdd_app/ui/sample_completion/metadata/MetadataFragment.kt
+++ b/app/src/main/java/net/aiscope/gdd_app/ui/sample_completion/metadata/MetadataFragment.kt
@@ -46,21 +46,29 @@ class MetadataFragment : SampleFormFragment, Fragment(R.layout.fragment_metadata
 
                 smearTypeId?.let { metadataSectionSmearTypeRadioGroup.check(it) }
                 speciesValue?.let { metadataSpeciesSpinner.select(it) }
+                comments?.let {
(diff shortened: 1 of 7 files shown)


app/src/main/java/net/aiscope/gdd_app/ui/sample_completion/metadata/MetadataFragment.kt
```diff
@@ -46,21 +46,29 @@ class MetadataFragment : SampleFormFragment, Fragment(R.layout.fragment_metadata
...
         }
+    }
+    override fun validateAndUpdateVM(): Boolean {
+        updateVM()
         //No validation on this tab
         return true
     }
 
     private fun onAddImageClicked() {
+        updateVM()
```
rkunal, line 70: This is the main change introduced. Saving without validating. 

Now to find an easy way for the model to handle its sanity, before sending to remote.

PR #311, opened 2022-01-10
Bump buildversion to android 12

jbors:


simtop:
I couldn't check it locally, everything looks right. Could you please check something before merging?  Search in the Project, click in Scope all Places and search for PendingIntent.FLAG and check if everything is using or has a case where they use FLAG_IMMUTABLE or MUTABLE, if not we have to find which library hasn't moved to Android 12 and write an issue or update.

simtop (approved):

PR #343, opened 2022-02-26
[165] Show loading icon on saving mask

jbors:


app/src/main/java/net/aiscope/gdd_app/ui/mask/MaskActivity.kt
```diff
@@ -159,6 +160,7 @@ class MaskActivity : AppCompatActivity(), MaskView, CaptureFlow {
     }
 
     override fun takeMask(maskName: String, onPhotoReceived: suspend (File?) -> Unit) {
+        binding.maskLoadingModal.isVisible = true
```
simtop, line 163: I see you make the mask visible, but when do you turn it off?

jbors: It get shown while we save the mask image and then we finish the activity (in `goToSampleCompletion`)

PR #344, opened 2022-03-19
Add test backend

jbors:


jbors:
Subsumed by https://github.com/theaiscope/GDD-app/pull/359

PR #375, opened 2022-06-22
178 save sample info in firebase store

rkunal:
Create/Update Sample collection in firestore whenever sample's metadata.json is stored in cloud firebase

Update: 
1. SampleCollection entity should always be dependent on Sample. I dont forsee it being used in isolation.
2. I have avoided any abstractions for Firestore related operations for now. As, I do not know how it would be used in future. Any architecture changes related to that are welcome and will be implemented as part of this PR.


rkunal:
1. Currently using metadata.json timestamp format eg, 2022-06-22T21:10:19+0530. Instead of firestore's timestamp data type. Need to change
2. sampleID field is added in-order to fetch and update the collection

app/src/main/java/net/aiscope/gdd_app/network/FirestoreUtil.kt
```diff
@@ -0,0 +1,15 @@
+package net.aiscope.gdd_app.network
+import com.google.firebase.firestore.FirebaseFirestore
+
+class FirestoreUtil {
```
simtop, line 4: Hello, why do you create a new instance for Firebase Firestore? You could use FirebaseRemoteStorage to store the sample, that has an instance of the Firebase FireStore

app/src/main/java/net/aiscope/gdd_app/model/SampleModels.kt
```diff
@@ -31,6 +33,17 @@ data class Sample(
     fun nextImageName(): String = "${id}_image_${captures.completedCaptureCount()}"
 
     fun nextMaskName(): String = "${id}_mask_${captures.completedCaptureCount()}"
+
+    fun buildSampleCollection() : SampleCollection {
```
simtop, line 37: you can one line it doing fun buildSampleCollection() = SampleCollection(
            createdOn = Timestamp(createdOn.time),
            uploadedBy = FirestoreUtil.FirestoreUtil.firestore?.
            collection("microscopist")?.document(microscopist),
            location = id,
            numberOfImages = captures.completedCaptureCount()
        )

app/src/main/java/net/aiscope/gdd_app/ui/sample_completion/SampleCompletionViewModel.kt
```diff
@@ -146,6 +146,7 @@ class SampleCompletionViewModel @Inject constructor(
                     microscopistRepository.load().copy(hasSubmitSampleFirstTime = true)
                 )
             }
+            remoteStorage.storeSampleCollection(storedSample)
```
simtop, line 149: Is this needed? Having it we will end up saving 2 samples each time, I think modifying the code just on top of this to accept the new Model will solve the issue.

simtop (changes requested):

app/src/main/java/net/aiscope/gdd_app/ui/sample_completion/SampleCompletionViewModel.kt
jbors: They end up in different locations though. This new 'storedsample' goes to https://console.firebase.google.com/u/0/project/aiscope-labelling-app-test/firestore/data/~2Fsamples~2F3a03bbd6-4047-4094-8c4a-3e6fa0863898

As opposed to https://console.firebase.google.com/u/0/project/aiscope-labelling-app-test/storage/aiscope-labelling-app-test.appspot.com/files/~2F3a03bbd6-4047-4094-8c4a-3e6fa0863898

app/src/main/java/net/aiscope/gdd_app/network/FirestoreUtil.kt
rkunal: I could not find firebaseFirestore instance in FirebaseRemoteStorage. FirebaseModule exists, but I am not sure how to use it. Unless I move all DB operations to SampleRepositorySharedPreference ?

PS: To understand things better, I have simplified the code, moved all DB operations to one SampleViewModel and commented everything else. Where should this code be best moved to ?

rkunal: @simtop  I have added a new repository to store sample in firestore. Currently it does not implement  samplerepository interface.  feel it's too early for that.  Cleaned up a bit. Let me know.

simtop (approved):

jbors (approved):

PR #382, opened 2022-07-02
Bump android-lint to latest version

jbors:


jbors:
So apparently this is not the bitrise.yml we actually use, we build with the configutation on bitrise

jbors:
Not relevant anymore

PR #498, opened 2024-02-05
Add coroutines dependency

jbors:


jbors:
Does not seem to fix our current issue but may still be a good idea.

@simtop What do you think?

PR #499, opened 2024-02-06
Reinstate removed package definition

jbors:


jbors:
Doesn't fix our issue, closing

PR #502, opened 2024-02-21
Downgrade all seemingly relevant parts

jbors:


jbors:
Subsumed by https://github.com/theaiscope/GDD-app/pull/508

PR #530, opened 2024-06-22
Updating auth firebase google services

simtop:
Updating google play services auth and firebase bom and  google services plugin.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -112,14 +112,14 @@ dependencies {
     implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-android:1.8.1'
 
     // Firebase
-    implementation platform("com.google.firebase:firebase-bom:32.7.3")
+    implementation platform("com.google.firebase:firebase-bom:33.1.1")
     implementation 'com.google.firebase:firebase-analytics'
     implementation 'com.google.firebase:firebase-crashlytics'
     implementation 'com.google.firebase:firebase-common-ktx'
     implementation 'com.google.firebase:firebase-storage-ktx'
     implementation 'com.google.firebase:firebase-firestore-ktx'
     implementation 'com.firebaseui:firebase-ui-auth:8.0.2'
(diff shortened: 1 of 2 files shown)


simtop:
Hello @jbors this worked from my side and all UI Tests passed locally, let's see if it works in bitrise.

simtop:
No, it's not fixed, the error comes between the virtual machine of bitrise and  R8 version and Kotlin compatibility, do you know if we made any changes related to that? @jbors 


"WARNING: D8: An error occurred when parsing kotlin metadata. This normally happens when using a newer version of kotlin than the kotlin version released when this version of R8 was created. To find compatible kotlin versions, please see: https://developer.android.com/studio/build/kotlin-d8-r8-versions
"

jbors:
I mean, yeah we bumped kotlin versions and I have been messing with proguard rules. Doesn't explain why it's only when we bump _this_ package that it goes wrong. Maybe it matters which version of kotlin the package has been compiled on?

simtop:
Yes probably, I tried lowering kotlin version, the problem is between bitrise virtual machine app and kotlin version, maybe they are behind and there is some incompatibility. 

I think the VM job passed, if everything passes let's merge it, and update to kotlin 2.0 when it's stable.

jbors:
I mean, alternatively we just do not update this for now and wait for kotlin version to catch up?

simtop:
Kotlin 2.0 is already out there but for now it only works for Android Studio canary versions, I think it's better to lower the kotlin version for now so it doesn't block us from updating anything else. El 22 jun 2024 10:28, Joris Borsboom ***@***.***> escribió:
I mean, alternatively we just do not update this for now and wait for kotlin version to catch up?

simtop:
@jbors Let's keep this branch/PR open to experiment when we can update few more things, it looks like it still fails. It's best to retry again when we update to kotlin 2.0

jbors:
@simtop Cool. Thanks for helping out!

